PR #1, opened 2020-07-17
Setup DID class

PaulLeCam:
Opening as draft for early feedback.


src/did.ts
```diff
@@ -0,0 +1,75 @@
...
+      jsonrpc: '2.0',
+      id: `did${this._reqId++}`,
+      method,
+      params,
+    })
+    if (res.error != null) {
+      throw new Error(res.error.message ?? 'RPC error')
+    }
+    return res.result as R
+  }
```
PaulLeCam, line 52: @oed I'm wondering if this request envelope logic should live in this class or somewhere else please?
IMO it's a provider/transport layer concern, but if the provider only takes care of implementing the `send: (request: RpcRequest) => Promise<RpcResponse>` interface I guess it has to live here?

oed: I think it can live here for now. We have some similar logic in `3id-blockchain-utils` and 3box-js. I'm not sure if there is one best place to put it.

PaulLeCam:
@oed should be ready for review now.

.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
+name: CI
+on: [push]
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
oed, line 5: Nice, so the CI runs on github actions now rather than CircleCI as our other libs? Probably simplifies things :+1:

README.md
```diff
@@ -2,3 +2,68 @@
...
+
+// Authenticate with the provider
+await alice.authenticate()
+
+// Read the DID string - this will throw an error if the DID instance is not authenticated
+const aliceDID = alice.DID
+
+// Create a JWS - this will throw an error if the DID instance is not authenticated
+const jws = await alice.createJWS({ hello: 'world' })
+```
```
oed, line 29: Neat! Quite clear example. Maybe we should include something that shows that objects that include CIDs are supported as well?

package.json
```diff
@@ -4,11 +4,11 @@
   "description": "Typescript library for interacting with DIDs",
   "main": "lib/index.js",
   "scripts": {
-    "test": "jest --coverage",
-    "build": "./node_modules/.bin/tsc -p tsconfig.json",
+    "test": "jest",
+    "build": "tsc -p tsconfig.json",
```
oed, line 8: Pretty sure this may cause issues if you don't have `tsc` installed gobaly?

README.md
PaulLeCam: 👍 Yes good idea

package.json
PaulLeCam: When using `npm run`, the `./node_modules/.bin` folder is added to the env's path, so it's unnecessary, cf https://docs.npmjs.com/cli/run-script

.github/workflows/main.yml
PaulLeCam: Yes I just used https://tsdx.io/ to scaffold the project and it came with this setup so that seemed like a good default to keep!

PR #5, opened 2020-08-19
Ensure provider can't be changed after it's been set

PaulLeCam:


src/index.ts
```diff
@@ -64,9 +64,10 @@ export class DID {
   }
 
   setProvider(provider: DIDProvider): void {
-    if (this._client?.connection !== provider) {
+    if (this._client == null) {
       this._client = new RPCClient(provider)
-      this._did = undefined
+    } else if (this._client.connection !== provider) {
+      throw new Error('A different provider is already set')
```
oed, line 70: Change to -> `A different provider is already set, create a new DID instance to use another provider`

oed (commented):
Small nitpick.

PR #6, opened 2020-08-20
feat: add createDagJWS method

oed:
Have not implemented tests yet. Not sure why linter is complaining (ignored the line).


src/index.ts
```diff
@@ -18,18 +18,24 @@ interface AuthenticateResult {
...
 export interface CreateJWSOptions {
+  did?: string
   protected?: Record<string, any>
+  linkedBlock?: string
 }
 
 interface CreateJWSParams extends CreateJWSOptions {
-  did: string
-  payload: DagJson
+  payload: string | any
```
PaulLeCam, line 27: Maybe we can use `Record<string, any>` rather than `any` if the payload is expected to be an object?

src/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+
+interface EncodedPayload {
+  cid: string
+  linkedBlock: string
+}
+
+export async function encodePayload(payload: Record<string, any>): Promise<EncodedPayload> {
+  const block = dagCBOR.util.serialize(payload)
+  return {
+    cid: base64url.encode(Buffer.from((await dagCBOR.util.cid(block)).bytes)),
```
PaulLeCam, line 12: I think we should avoid using `Buffer` if we can, especially since IFPS seems to replace them by the standard ArrayBuffer.

src/index.ts
oed: Payload can really be anything that is valid JSON.

src/utils.ts
oed: I agree, is there a better way to do this? `base64url.encode` requires Buffer or string as type, but it **does** work with an Uint8Array. Linter complains however.

src/index.ts
PaulLeCam: 👍 we could just use `any` then as `string` is included in `any`

oed: Sure!

src/utils.ts
PaulLeCam: Maybe we can have our own utility function doing the encoding?
Using modern browser APIs it seems pretty simple, ex: https://github.com/denoland/deno/blob/master/std/encoding/base64.ts and https://github.com/denoland/deno/blob/master/std/encoding/base64url.ts

I think the encoding function would be something like this (not tested):
```ts
function encodeBase64URL(data: Uint8Array): string {
  let dataString = ''
  for (let i = 0; i < data.length; ++i) {
    dataString += String.fromCharCode(data[i])
  }

  return btoa(dataString)
    .replace(/=/g, '')
    .replace(/\+/g, '-')
    .replace(/\//g, '_')
}
```

oed: Looks like btoa can't be used in nodejs?
For reference:
```js
function encodeb64(data: Uint8Array): string {
  const base64str = btoa(data.map(String.fromCharCode).join(''))
  return base64str
    .replace(/=/g, '')
    .replace(/\+/g, '-')
    .replace(/\//g, '_')
```

PaulLeCam: Ah yes sorry, I forgot it's a DOM API.
Back to using `base64url` then, maybe we can still avoid using `Buffer` and just add a `ts-ignore` comment if the types don't match?

oed: I tested it again and it doens't work with `Uint8Array`. base64url depends on buffer to stringify as base64 :/

oed:
Thanks for reminding me @PaulLeCam Added an update to the readme!

PR #11, opened 2020-09-21
Feat/encryption

oed:
Adds support for encryption and decryption using JWEs.
Adds documentation generated with typedoc


src/utils.ts
```diff
@@ -15,11 +21,34 @@ export async function encodePayload(payload: Record<string, any>): Promise<Encod
...
+export function decodeBase64(s: string): Uint8Array {
+  return u8a.fromString(s, B64)
+}
+
+export function decodeBase58(s: string): Uint8Array {
+  return u8a.fromString(s, B58)
 }
 
-export interface JWSSignature {
+interface JWSSignature {
```
PaulLeCam, line 51: Can you make sure it's exported please? I'm using this interface in `idx-tools`

src/index.ts
```diff
@@ -1,11 +1,21 @@
...
+  encodeCleartext,
+  decodeCleartext,
+  encodeBase64,
+  decodeBase64,
+  decodeBase58,
+} from './utils'
 
 export type { DIDDocument } from 'did-resolver'
 
-export type { DagJWS, JWSSignature } from './utils'
```
PaulLeCam, line 8: Can you make sure it's exported please? I'm using this interface in `idx-tools`

PaulLeCam (approved):
Nice!
I just noticed the removal of a type export that should be kept if possible please?

oed:
@PaulLeCam Exports added back.

PaulLeCam:
Thanks!

zachferland (approved):
🎉 👍 

the worst diff, the docs could just be pushed to s3 in the future right?

oed:
Good point @zachferland There is probably even some github action to do that. 
cc @simonovic86

PR #12, opened 2020-09-22
Fix/add padding

oed:
Totally forgot that we needed padding. Here it is.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,6 +5,8 @@ import * as u8a from 'uint8arrays'
 
 const DAG_CBOR_CODE = 113
 const ID_MULTIHASH = 0
+const ENC_BLOCK_SIZE = 24
+const PAD_FIRST_BYTE = 128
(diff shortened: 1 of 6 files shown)


src/utils.ts
```diff
@@ -21,13 +22,28 @@ export async function encodePayload(payload: Record<string, any>): Promise<Encod
   }
 }
 
-export function encodeCleartext(cleartext: Record<string, any>): CID {
+export function pad(b: Uint8Array, blockSize = ENC_BLOCK_SIZE): Uint8Array {
+  // pads with 1 followed by 0s
+  const padLen = blockSize - (b.length % blockSize)
+  return u8a.concat([b, new Uint8Array([128]), new Uint8Array(padLen - 1)])
```
PaulLeCam, line 28: Not an issue but maybe `128` could be set as a constant to avoid having "magic numbers" in functions?

PaulLeCam (approved):

PR #13, opened 2020-09-25
Fix encoding

simonovic86:
- change `encodeBase64` to `encodeBase64Url `

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -9,6 +9,7 @@ const ENC_BLOCK_SIZE = 24
 const PAD_FIRST_BYTE = 128
 const B64 = 'base64pad'
 const B58 = 'base58btc'
+const B64_URL = 'base64urlpad'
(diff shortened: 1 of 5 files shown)


PaulLeCam (commented):
LGTM but I'll leave it to @oed to confirm the expected behavior.

src/utils.ts
```diff
@@ -9,6 +9,7 @@ const ENC_BLOCK_SIZE = 24
 const PAD_FIRST_BYTE = 128
 const B64 = 'base64pad'
 const B58 = 'base58btc'
+const B64_URL = 'base64urlpad'
```
oed, line 12: base64url used in JOSE doesn't use padding.
```suggestion
const B64_URL = 'base64url'
```

oed (commented):
Looks good, just one small change needed.

PR #15, opened 2020-10-21
Feat/add jws verification method

oed:
Adds a `verifyJWS` methods that verifies a JWS based on the `kid` in the header.


PaulLeCam (approved):
Nice!

PR #17, opened 2020-11-13
Release/0.8.0

oed:
* Bump did-jwt to support key-did-provider-ed25519
* Future proof the returned result  from `verifyJWS`
* Generate docs


PaulLeCam (approved):
Could we move the generated docs to another branch so these diffs don't have to be included?
Using the [`gh-pages` CLI](https://www.npmjs.com/package/gh-pages#command-line-utility) it should be a simple setup.

oed:
Makes sense to use gh-pages. Can we have it automatically build to that branch using an action?

PaulLeCam:
Seems GitHub actions could support this yes, but I'm not familiar enough to know the full process.

PR #25, opened 2021-03-10
feat!: Use did-resolver v3

oed:
This is a breaking change because the api of the `did-resolver` package has changed. Namely what the format of the result object looks like and the types used for the resolver registry.

Also generated the docs.


src/index.ts
```diff
@@ -217,18 +217,17 @@ export class DID {
     if (typeof jws !== 'string') jws = fromDagJWS(jws)
     const kid = base64urlToJSON(jws.split('.')[0]).kid as string
     if (!kid) throw new Error('No "kid" found in jws')
-    const { publicKey } = await this.resolve(kid)
+    const didResolutionResult = await this.resolve(kid)
+    const publicKeys = didResolutionResult.didDocument!.verificationMethod || []
```
PaulLeCam, line 221: As noted by the typecheck warning below, it seems unsafe as `didDocument` could be `null` according to this type: https://github.com/decentralized-identity/did-resolver/blob/master/src/resolver.ts#L17

PaulLeCam (commented):
LGTM apart from the typecheck warning raised, any reason it is ignored?

src/index.ts
oed: It will really only be `null` if there is an error, which is caught in the `this.resolve` method. Maybe I can improve it there?

oed: Updated to use `?` instead.

PaulLeCam (approved):

PR #28, opened 2021-04-06
Split authenticated and resolvable interfaces

PaulLeCam:
@oed @stbrody following on https://github.com/ceramicnetwork/js-ceramic/pull/1199 I wondered if the problem with circular dependencies with 3ID DID provider could be coming from the DID implementation that does both DID provider and DID resolver interactions?

Here I split the logic between the "authenticated" (that requires a DID provider) and "resolvable" (that requires a DID resolver) interfaces and functions, so then the only case we really need the DID instance to handle both is to authenticate.
Do you think instead of using the DID instance directly in Ceramic, the APIs could use these interfaces so maybe that would break the circle of dependencies somehow, or at least give us a better understanding of how to work around it?

The code in this draft is not great, mostly POC to present the idea so we can discuss the implementation and naming further if we want to go in this direction.


stbrody:
I had a similar thought myself recently.  I think we'll need to workshop the names and exact APIs a little, but I'm a fan of the overall direction.  Making the two separate roles of a `DID` instance explicit (resolving arbitrary DIDs vs authenticating a specific DID) and allowing code the specify which pieces it needs where seems quite valuable.

src/resolvable.ts
```diff
@@ -0,0 +1,74 @@
+import type { DIDResolutionResult, Resolver } from 'did-resolver'
+import { createJWE, verifyJWS, resolveX25519Encrypters } from 'did-jwt'
+import type { JWE } from 'did-jwt'
+import { prepareCleartext } from 'dag-jose-utils'
+
+import { fromDagJWS, base64urlToJSON } from './utils'
+import type { DagJWS } from './utils'
+
+export interface Resolvable {
```
stbrody, line 9: Instead of a new `Resolvable` interface, could we-use the types from [did-resolver](https://github.com/decentralized-identity/did-resolver/blob/master/src/resolver.ts)?  Maybe the `DID` class could extend `Resolver`?  That way you can ask for a `Resolver` when that's the only functionality you need, but could provide a `DID` instance directly without having to reach into it for its underlying `Resolver`?

PaulLeCam: Yeah I didn't know did-resolver already had an interface for it, so I updated the DID class to implement this interface 👍

PaulLeCam:
I updated the logic a bit to have a `Providable` interface and `Authenticated` one extending it with strict `authenticated` and `id` properties constraints.
The idea is that low-level APIs should just accept `Authenticated` inputs, but higher-level ones can accept `Providable` ones and call `toAuthenticated()`. Essentially `Providable` allows for lazy authentication when needed, while `Authenticated` should provide the guaranty that authentication has already been performed.

I added the `AuthDID` class extending `DID` but also satisfying `Authenticated` constraints so it should be possible to use `AuthDID` as `Authenticated` input.

PaulLeCam:
Yeah I agree, I don't think we need to release these changes before the ELP but if there are interfaces being created or changed in Ceramic before the ELP would be good to keep this in sync so we don't end-up making too many breaking changes after the ELP.

PR #32, opened 2021-06-21
feat: Check for revoked keys when verifying JWS

ukstv:


src/did.ts
```diff
@@ -0,0 +1,321 @@
...
+   * Verify a JWS. Uses the 'kid' in the header as the way to resolve
+   * the author public key.
+   *
+   * @param jws                 The JWS to verify
+   * @param atTime              JS timestamp when to verify the key
+   * @returns                   Information about the signed JWS
+   */
+  async verifyJWS(
+    jws: string | DagJWS,
+    atTime: number = new Date().valueOf()
```
oed, line 211: I would pass an options object here since that gives us more flexibility if we need to other params in the future.

src/did.ts
```diff
@@ -0,0 +1,321 @@
...
+        // Key is definitely revoked
+        const revokedAt = new Date(
+          nextDidResolutionResult.didDocumentMetadata.created || '0'
+        ).valueOf()
+        if (atTime >= revokedAt) {
+          // Do not allow using a key _after_ it is being revoked
+          throw new Error(`Key is revoked`)
+        }
+      }
+    }
```
oed, line 234: This seems overcomplicated and we should definitely not need to resolve the DID twice (that was why I spent a bunch of time adding this new feature to *did-core*). The suggestion should suffice!
```suggestion
    const nextUpdate = didResolutionResult.didDocumentMetadata?.nextUpdate
    if (nextUpdate) {
      // This key has been revoked. Check if the JWS 
      // was signed before it the revocation happened.
      if (atTime >= new Date(nextUpdate).valueOf()) {
       // Do not allow using a key _after_ it is being revoked
        throw new Error(`JWS was signed with a revoked key: ${kid}`)
      }
    }
```

oed (changes requested):
We can simplify the logic a lot and only do one DID resolution!

src/did.ts
```diff
@@ -0,0 +1,321 @@
+import {
+  ResolverOptions,
```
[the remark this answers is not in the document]
ukstv: I guess no, but _no_ changes have been made to methods other than `verifyJWS`.

src/did.ts
ukstv: Is it _always_ the case, that an update to a DID document (just in case, we do not know what method is used here) results in the keys being revoked? If a DID document is updated, but it adds a key, or removes other key [and leaves the previously used signing key intact], the less-complicated design would report false failure.

src/did.ts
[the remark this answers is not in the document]
oed: ok, should be fine then.

src/did.ts
oed: While the key *could* still be contained in the next version of the document, the `kid` that was specified is still invalid and we should consider the signature invalid imo. Otherwise for a really old version of the DID document where a key was valid until the last update we would have to resolve **all** updates inbetween the old and the last update which would be extremely inefficient.

ukstv: Okay, I consider this to be a simplification, that would work in 90% of cases.

oed: It should work in 100% of the cases. DID providers should make sure to always include the current version-id/time when creating JWSs

ukstv: Yes, but availability of the next version does not necessarily mean, the actual key has been revoked.

oed: While that's true, the assertment we are making is that the JWS was signed with the given `kid` if the DID document version is no longer valid then we should consider the JWS invalid.

oed: Updated the error thrown to reflect this.

package.json
```diff
@@ -19,9 +19,6 @@
     "type": "git",
     "url": "git+https://github.com/ceramicnetwork/js-did.git"
   },
-  "jest": {
-    "testEnvironment": "jest-environment-uint8array"
```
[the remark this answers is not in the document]
ukstv: Removing that does not break tests.

src/__tests__/jws-verification-behavior.test.ts
```diff
@@ -0,0 +1,183 @@
...
+    const isVersion0 = /version-id=0/.exec(didUrl)
+    if (isVersion0) {
+      return Promise.resolve(VERSION_0_ROTATED)
+    } else {
+      return Promise.resolve(VERSION_NEXT)
+    }
+  }
+
+  const jws =
+    'eyJraWQiOiJkaWQ6MzpranpsNmN3ZTFqdzE0YWNzeHRvMHd2eHdsNjJ1c3EzMnk0bmNmb256ZG1nNWk1am9yamNzMmhhbml1YWJ2emw_dmVyc2lvbi1pZD0wI2RrYm5CM0RkMXo1d0J5biIsImFsZyI6IkVTMjU2SyJ9.eyJoZWxsbyI6IndvcmxkIn0._FMNUc83qOfNEyOUvpQs09CX12MbLEa90BIqa6qLw96VT2pec26gdlin5ozJ4veVgLbRrUBFCUUP2iLn6JtPhQ'
```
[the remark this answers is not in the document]
ukstv: yup

src/did.ts
ukstv: kk

src/did.ts
ukstv: changed.

oed (approved):
lgtm 🎉

PR #34, opened 2021-07-02
feat: check if a signature made earlier using implicit _now_.

ukstv:


oed (approved):
lgtm

src/did.ts
```diff
@@ -208,15 +208,15 @@ export class DID {
...
-    const effectiveOptions = Object.assign({ atTime: new Date().valueOf() }, options)
     const kid = base64urlToJSON(jws.split('.')[0]).kid as string
     if (!kid) throw new Error('No "kid" found in jws')
     const didResolutionResult = await this.resolve(kid)
     const nextUpdate = didResolutionResult.didDocumentMetadata?.nextUpdate
     if (nextUpdate) {
       // This version of the DID document has been revoked. Check if the JWS
       // was signed before it the revocation happened.
-      if (effectiveOptions.atTime >= new Date(nextUpdate).valueOf()) {
+      const isEarlier = options.atTime && options.atTime < new Date(nextUpdate).valueOf()
```
stbrody, line 218: Didn't we discuss in the other PR that we also want to throw if `nextUpdate` is set but `atTime` is undefined?  @oed @ukstv

ukstv: If atTime is `undefined`, then `isEarlier` is `undefined` as well, which makes `!isEarlier` become `true`, and it will throw.

PR #36, opened 2021-07-05
feat: Disallow using a key before it is created

ukstv:
Unless it is v0, which can be created whenever.

### DIFF
--- a/src/__tests__/__snapshots__/jws-verification-behavior.test.ts.snap
+++ b/src/__tests__/__snapshots__/jws-verification-behavior.test.ts.snap
@@ -1,3 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`atTime before DID created for v0 1`] = `"did:3:kjzl6cwe1jw14acsxto0wvxwl62usq32y4ncfonzdmg5i5jorjcs2haniuabvzl?version-id=0#dkbnB3Dd1z5wByn"`;
+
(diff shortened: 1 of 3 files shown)


oed (changes requested):
Why is this change needed exactly? I don't think we should make such a check at all
We should definitely not have code in here that is specific to 3ID!

ukstv:
To address a scenario from here https://github.com/ceramicnetwork/js-ceramic/pull/1518#discussion_r663119353

Maybe it is impossible for 3id, but it feels like it can be the case for other DID methods.

The only thing that links to 3id is `versionId: 0`. The only reason to have that, IMO, is 3id resolution incorrectly returns `created` timestamp for the genesis commit as the timestamp of the first anchor commit, which goes against the DID Core specification (again, in my opinion).

[DID Core Specification](https://www.w3.org/TR/did-core/#did-document-metadata): "DID document metadata SHOULD include a created property to indicate the timestamp of the Create operation"

For real 3id "create operation" can happen any time really. A genesis commit is not anchored by itself. Only updates are anchored. So, if we remove `created` field from resolved 3id did document v0, this whole check stops being 3id-specific.

stbrody:
Where did we end up on this? @ukstv @oed?

ukstv:
This one is closed: it is too 3id-specific. There is another one based on "updated" property.

stbrody:
> This one is closed: it is too 3id-specific. There is another one based on "updated" property.

Is that this one: https://github.com/ceramicnetwork/js-ceramic/pull/1524?

ukstv:
> > This one is closed: it is too 3id-specific. There is another one based on "updated" property.
> 
> Is that this one: [ceramicnetwork/js-ceramic#1524](https://github.com/ceramicnetwork/js-ceramic/pull/1524)?

yes, it is.

PR #39, opened 2021-07-06
chore: Disallow using a key before it is created: based on update property

ukstv:


src/did.ts
```diff
@@ -216,11 +216,19 @@ export class DID {
...
+      if (isLater) {
         // Do not allow using a key _after_ it is being revoked
         throw new Error(`JWS was signed with a revoked DID version: ${kid}`)
       }
     }
+    // Key used before `updated` date
+    const updated = didResolutionResult.didDocumentMetadata?.updated
+    const versionId = didResolutionResult.didDocumentMetadata?.versionId
+    const notV0 = versionId !== '0'
+    if (notV0 && updated && options.atTime && options.atTime < new Date(updated).valueOf()) {
```
oed, line 229: Were we not going to remove the versionId check here, as it's 3ID specific?

oed: Also what happens if updated is not defined here. Can be the case for various DID methods, e.g. Key DID.

ukstv: Ooh. Stupid me. Removed that in https://github.com/ceramicnetwork/js-did/pull/40

ukstv: > Also what happens if updated is not defined here. Can be the case for various DID methods, e.g. Key DID.

No `updated` ⇒ no reference timestamp ⇒ do not throw.

oed: I see, silly me :P

ukstv:
So @oed is it a formal yes to merge both PRs?

oed (approved):
Approved on the condition that #40 get's merged ;)

PR #41, opened 2021-07-12
feat: Pass issuer to verifyJWS

ukstv:
If `did-A` is the controller of `did-B` a signature performed by A should be seen as valid if B is the issuer. We use the controller property in DID documents to allow for Safe and NFT DIDs.

### DIFF
--- a/src/did.ts
+++ b/src/did.ts
@@ -251,7 +251,7 @@ export class DID {
       const controllers = extractControllers(controllerProperty)
       const signerIsController = controllers.some((controller) => controller === signerDid)
       if (!signerIsController) {
-        throw new Error(`invalid_jws: kid does not belong to issuer`)
+        throw new Error(`invalid_jws: not a valid verificationMethod for issuer: ${kid}`)
       }
     }
 
--- a/src/did.ts
+++ b/src/did.ts
@@ -249,7 +249,7 @@ export class DID {
       const issuerResolution = await this.resolve(issuerUrl)
       const controllerProperty = issuerResolution.didDocument?.controller
       const controllers = extractControllers(controllerProperty)
-      const signerIsController = controllers.some((controller) => controller === signerDid)
+      const signerIsController = signerDid ? controllers.includes(signerDid) : false
       if (!signerIsController) {
         throw new Error(`invalid_jws: not a valid verificationMethod for issuer: ${kid}`)
       }
--- a/src/did.ts
+++ b/src/did.ts
@@ -11,6 +11,8 @@ import {
   decodeBase64,
   encodeBase64Url,
   randomString,
+  didWithTime,
+  extractControllers,
 } from './utils'
 
 export interface AuthenticateOptions {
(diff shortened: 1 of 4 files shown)


src/did.ts
```diff
@@ -236,6 +243,18 @@ export class DID {
       }
     }
 
+    const signerDid = didResolutionResult.didDocument?.id
+    if (options.issuer && options.issuer !== signerDid) {
+      const issuerUrl = didWithTime(options.issuer, options.atTime)
+      const issuerResolution = await this.resolve(issuerUrl)
+      const controllerProperty = issuerResolution.didDocument?.controller
+      const controllers = extractControllers(controllerProperty)
+      const signerIsController = controllers.some((controller) => controller === signerDid)
```
oed, line 252: Why not `.includes`?

src/did.ts
```diff
@@ -236,6 +243,18 @@ export class DID {
...
+      if (!signerIsController) {
+        throw new Error(`invalid_jws: kid does not belong to issuer`)
```
oed, line 254:
```suggestion
        throw new Error(`invalid_jws: not a valid verificationMethod for issuer: ${kid}`)
```

src/did.ts
```diff
@@ -236,6 +243,18 @@ export class DID {
       }
     }
 
+    const signerDid = didResolutionResult.didDocument?.id
+    if (options.issuer && options.issuer !== signerDid) {
+      const issuerUrl = didWithTime(options.issuer, options.atTime)
```
oed, line 248: Maybe break this and the `timecheckEnabled` conditional into helper functions?

oed (changes requested):
Looks good, suggesting a few changes.

src/did.ts
ukstv: TBH, I would like to accumulate a little more weirdness like this, and extract `verifyJWS` into own verification service. It clearly does not belong here as a method of `DID` object.

src/did.ts
ukstv: Forgot about it :)

src/did.ts
oed: Well, ideally most of this logic could move into the did-jwt package. Which btw should be renamed did-jose at some point: https://github.com/decentralized-identity/did-jwt/issues/170

ukstv: Ideally, yeah.

src/did.ts
```diff
@@ -249,7 +249,7 @@ export class DID {
       const issuerResolution = await this.resolve(issuerUrl)
       const controllerProperty = issuerResolution.didDocument?.controller
       const controllers = extractControllers(controllerProperty)
-      const signerIsController = controllers.some((controller) => controller === signerDid)
+      const signerIsController = signerDid ? controllers.includes(signerDid) : false
```
oed, line 252: eh, ok I see why now 😆

src/did.ts
ukstv: I really want to accumulate weirdness here nevertheless to guide the further changes.

src/did.ts
ukstv: uhums :)

oed (approved):

PR #53, opened 2021-11-24
feat: Expose pure ESM

ukstv:
Apply ESM-transition checklist.

Divergence:
- have to add tsconfig.build.json to not include test files to NPM package.

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -1,3 +1,3 @@
 {
-  "presets": [["@babel/preset-env", { "targets": { "node": 14 } }], "@babel/preset-typescript"]
+  "presets": [["@babel/preset-env", { "targets": { "node": 16 } }], "@babel/preset-typescript"]
 }
--- a/babel.config.json
+++ b/babel.config.json
@@ -1,0 +1,3 @@
+{
+  "presets": [["@babel/preset-env", { "targets": { "node": 14 } }], "@babel/preset-typescript"]
+}
--- a/package.json
+++ b/package.json
@@ -12,12 +12,12 @@
   ],
   "sideEffects": false,
   "scripts": {
-    "test": "jest",
+    "test": "./node_modules/.bin/jest",
     "build": "./node_modules/.bin/tsc --project tsconfig.build.json",
     "prepublishOnly": "npm run build",
     "prebuild": "npm run clean",
-    "lint": "eslint ./src --ext .js,.ts",
-    "docs": "typedoc ./src/index.ts --excludePrivate --name DID --includeVersion ; touch docs/.nojekyll",
+    "lint": "./node_modules/.bin/eslint ./src --ext .js,.ts",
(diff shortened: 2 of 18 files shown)


babel.config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "presets": [["@babel/preset-env", { "targets": { "node": 14 } }], "@babel/preset-typescript"]
```
PaulLeCam, line 2: Should we target node 16?

package.json
```diff
@@ -2,13 +2,18 @@
...
+    ".": "./lib/index.js"
+  },
   "files": [
     "lib/*"
   ],
+  "sideEffects": false,
   "scripts": {
     "test": "jest",
-    "build": "tsc -p tsconfig.json",
+    "build": "./node_modules/.bin/tsc --project tsconfig.build.json",
```
PaulLeCam, line 16: Is this needed to specify the path to `tsc`? My understanding is that npm handles this, no?

PaulLeCam (approved):
Nice!

babel.config.json
ukstv: Fair point.

package.json
ukstv: `npm` handles this indeed by providing `$PATH` that (for npm!) starts with `./node_modules/.bin`. A user though could forget to `npm install`: global (and it is another question what is "global" based on `$PATH` provided by npm, yarn or pnpm) `tsc` will be used. Or, sometimes, you could decide to run a script without relying on npm `$PATH`. Then you have to convert path to the executable by yourself. I understand, that these are edge cases, but they are really painful, so unless you strongly oppose this change, I would like to go with `./node_modules/.bin/...`.

PaulLeCam: I'm wondering how likely are these edge cases to happen? In CI this should never be the case unless there's a config issue, and locally I'd expect forgetting to run `npm install` would be problematic anyways.

I don't mind having this change if you feel strongly about it, I guess it's mostly that I'd like to have something consistent between scripts and repos, so here if we do this for `tsc` I think we should also do it for `eslint`, `jest` and any other, no?
It's a bit annoying though if we start moving to a monorepo (not in scope now but I think we could have this package, the key DID resolver and providers in the same repo), where I think we'd probably want to use the hoisted dependencies instead.

ukstv: Ooh, if moving to a monorepo is on the agenda (even in future), then using paths instead of command names is a necessity. I had a lot of confusion/mismatches in js-ceramic repository. Going to add paths to jest, eslint and others.

Just in case, agree on low expected probability of the edge cases.

ukstv:
Ping @stephhuynh18 Thanks for finding node v14 in GitHub action config. Changed to v16.

ukstv:
May I have a second approval, pwease, to conform to 50%+ approve rule? :P @stephhuynh18 @oed

PR #55, opened 2022-01-04
add cacao verification to verifyJWS

haardikk21:
# [Add CACAO Signature verification to verifyJWS]

## Description

Adds verification of optional `capability` CACAO OCAP in `verifyJWS`

## How Has This Been Tested?

Added a new test under `provider-behavior.test.ts` which tests verifying a JWS with a capability attached.

## Definition of Done

- [ ] I have made corresponding changes to the documentation

## References:

### DIFF
--- a/src/__tests__/provider-behavior.test.ts
+++ b/src/__tests__/provider-behavior.test.ts
@@ -262,9 +262,9 @@ describe('`createDagJWS method`', () => {
               signatures: [
                 {
                   protected:
-                    'eyJhbGciOiJFZERTQSIsImNhcCI6ImlwZnM6Ly9iYWZ5cmVpYm1oam1xdGR2d3hoZm03aXdwY2p6dWJ6anJuNmI0dWQ0dXd0Y3lkcXptaWZ0c3FmcXlyZSIsImtpZCI6ImRpZDprZXk6ejZNa21XWEplVllVbUFHUUhqZWlIYkpuMXpXMVZ1TXpQNFFlZnV4WjVpZUI3eEFyI3o2TWttV1hKZVZZVW1BR1FIamVpSGJKbjF6VzFWdU16UDRRZWZ1eFo1aWVCN3hBciJ9',
+                    'eyJhbGciOiJFZERTQSIsImNhcCI6ImlwZnM6Ly9iYWZ5cmVpY2szajZ2d2E3bTJ5dWgzb3VuY3ZkM2x1cTRtMzNlc3Q3NWU0dWQ3emx1azZza2V4eXN6NCIsImtpZCI6ImRpZDprZXk6ejZNa3JCZE5kd1VQblhEVkQxREN4ZWR6VlZCcGFHaThhU21vWEZBZUtOZ3RBZXI4I3o2TWtyQmROZHdVUG5YRFZEMURDeGVkelZWQnBhR2k4YVNtb1hGQWVLTmd0QWVyOCJ9',
                   signature:
-                    'eStPWv8KzmSCFSvHoxEmaJJ34TDsJh15OfOdFCI1bjE8RxvQbNU6Cg-nx0CcWz6cp-ZfOa5dre4fTpw4N49cBg',
+                    'GpxlcwZ8ouPVV4rYVxc5fka6-9rq4iVKwRLejh7IdVoa50gqP8wn01ivlv4PlpITNJUBELKqSZhUaWnG9wg5Dg',
                 },
               ],
             },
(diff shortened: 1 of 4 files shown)


haardikk21:
unsure why linting is failing - passes on local

src/did.ts
```diff
@@ -304,9 +308,22 @@ export class DID {
       const issuerResolution = await this.resolve(issuerUrl)
       const controllerProperty = issuerResolution.didDocument?.controller
       const controllers = extractControllers(controllerProperty)
-      const signerIsController = signerDid ? controllers.includes(signerDid) : false
-      if (!signerIsController) {
-        throw new Error(`invalid_jws: not a valid verificationMethod for issuer: ${kid}`)
+
+      if (options.capability && options.capability.s) {
+        if (
+          (options.capability.p.iss === options.issuer ||
```
oed, line 314: Can we please check this condition first before we resolve the `issuerURL` DID resolution is expensive and with the current implementation we will always resolve the DID even if the cacao iss is equal to the options.issuer.

src/__tests__/provider-behavior.test.ts
```diff
@@ -295,31 +364,30 @@ describe('`createDagJWS method`', () => {
     const res = await did.createDagJWS(data)
     const encPayload = await encodePayload(data)
 
+    expect(async () => {
+      await did.verifyJWS(res.jws, {
+        issuer: `did:pkh:eip155:1:${wallet.address}`,
+        capability: cacao,
+        atTime: Date.parse('2021-10-30T16:25:24.000Z'),
```
oed, line 371: The `atTime` needs to be a unix timestamp still, as. that's what Ceramic will pass: https://github.com/ceramicnetwork/js-ceramic/blob/1b4465167b70292145657fc5e272b7e79effeb2c/packages/common/src/stream.ts#L115

Cacao needs to convert this to an RFC data internally to do the comparison.

oed (changes requested):
Ceramic will pass the `atTime` variable as a UNIX timestamp, so the api needs to account for that.

src/__tests__/provider-behavior.test.ts
haardikk21: Date.parse returns a unix timestamp from a string

haardikk21: the internal repr in Cacao will be handled in the PR there, shouldnt affect the working of this i think

src/__tests__/provider-behavior.test.ts
```diff
@@ -270,23 +291,71 @@ describe('`createDagJWS method`', () => {
...
+              'did:key:z6MkmWXJeVYUmAGQHjeiHbJn1zW1VuMzP4QefuxZ5ieB7xAr#z6MkmWXJeVYUmAGQHjeiHbJn1zW1VuMzP4QefuxZ5ieB7xAr',
+            ],
+            keyAgreement: [
+              {
+                id: 'did:key:z6MkmWXJeVYUmAGQHjeiHbJn1zW1VuMzP4QefuxZ5ieB7xAr#z6LSrF5fu2BTRAxk4N6zPJSbc6ouoP1xRrvDX6FJJgr7Dzvx',
+                type: 'X25519KeyAgreementKey2019',
+                controller: 'did:key:z6MkmWXJeVYUmAGQHjeiHbJn1zW1VuMzP4QefuxZ5ieB7xAr',
+                publicKeyBase58: 'FZuWNiNbKiEzxyjDreveHWbRxEUqjFk4e7XcpECaWdAC',
+              },
+            ],
```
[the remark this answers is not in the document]
haardikk21: we need it for DID resolution because verifyJWS will resolve the `uri` / `cacao.aud` DID

src/__tests__/provider-behavior.test.ts
oed: Well, it returns the UNIX timestamp in ms, while Ceramic will pass the UNIX timestamp in seconds.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse

src/__tests__/provider-behavior.test.ts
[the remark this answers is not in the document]
oed: I'm just saying that we can remove some specific properties in the DID document since they are not used.

src/__tests__/provider-behavior.test.ts
haardikk21: Ah, yes. Should I just divide by 1000?

src/__tests__/provider-behavior.test.ts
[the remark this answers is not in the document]
haardikk21: Ah I see, okay

src/__tests__/provider-behavior.test.ts
oed: Actually I might be wrong here. Seems like we might actually use ms internally in Ceramic: https://github.com/ceramicnetwork/js-did/blob/main/src/did.ts#L296

oed: Investigating...

haardikk21: Yep! `valueOf` returns timestamp in ms. So I guess this is fine, and I will change internal workings of CACAO to use the RFC in a PR there.

oed: No, I'm trying to figure out what js-ceramic does. The code I linked from js-did might be a bug.

haardikk21: https://github.com/ceramicnetwork/js-ceramic/blob/1b4465167b70292145657fc5e272b7e79effeb2c/packages/stream-caip10-link-handler/src/__tests__/caip10-link-handler.test.ts#L55 these tests seem to suggest it is in seconds.

haardikk21: https://github.com/ceramicnetwork/js-ceramic/blob/1b4465167b70292145657fc5e272b7e79effeb2c/packages/stream-tile-handler/src/__tests__/tile-document.test.ts#L106 this too, as `commitData.timestamp` is set equal to `proof.blockTimestamp` in a couple of places

oed:
We get the timestamp from the anchor proof: https://github.com/ceramicnetwork/js-ceramic/blob/1b4465167b70292145657fc5e272b7e79effeb2c/packages/core/src/utils.ts#L97

Proof timestamp should be equal to block timestamp: https://github.com/ceramicnetwork/js-ceramic/blob/cdc47d3cd4156ab578f29e029c6fa1d84c552d98/packages/core/src/anchor/ethereum/ethereum-anchor-validator.ts#L158

The blocktimestamp from the ethereum API I assume is in seconds.

In memory anchor service also uses seconds: https://github.com/ceramicnetwork/js-ceramic/blob/28cfd622e684b3b7209884024e684be6e6a1fa88/packages/core/src/anchor/memory/in-memory-anchor-service.ts#L316

src/__tests__/provider-behavior.test.ts
haardikk21: It makes sense, block timestamp is always in seconds. `js-did` might be buggy after all.

oed:
@ukstv can you please have a look at this? Seems like we might have found a bug in the key revocation logic. js-did seems to be using ms while js-ceramic uses seconds.

Here: https://github.com/ceramicnetwork/js-did/blob/main/src/did.ts#L287
and here: https://github.com/ceramicnetwork/js-did/blob/main/src/did.ts#L296
`Date.valueOf()`would return timestamp in ms.

src/__tests__/provider-behavior.test.ts
```diff
@@ -270,56 +291,86 @@ describe('`createDagJWS method`', () => {
...
     }
 
     const res = await did.createDagJWS(data)
     const encPayload = await encodePayload(data)
 
+    expect(async () => {
+      await did.verifyJWS(res.jws, {
+        issuer: `did:pkh:eip155:1:${wallet.address}`,
+        capability: cacao,
+        atTime: Date.parse('2021-10-30T16:25:24.000Z'),
```
oed, line 354: Looks like we still are passing ms and not seconds.

oed (approved):

PR #65, opened 2022-02-09
Main import should point to module file

PaulLeCam:
## Description

Looks like the package.json setup is invalid, having a `main` file reference as CommonJS while the package is set as `module`.
This notably breaks imports with Jest.

## How Has This Been Tested?

Local change fixes test runner.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "3.0.0-alpha.6",
   "description": "Typescript library for interacting with DIDs",
   "type": "module",
-  "main": "./lib/index.cjs",
+  "main": "./lib/index.mjs",
   "types": "./lib/index.d.ts",
   "exports": {
     ".": {


ukstv:
@PaulLeCam I tend to believe `main` should reference CJS, as a "legacy" module-inclusion mechanism.

First, some random 2019 blog entry: https://2ality.com/2019/10/hybrid-npm-packages.html
Then, here is the proof it is used for CJS resolution: https://nodejs.org/api/modules.html#all-together
And, it only is used as a fallback for ESM if `exports` query fails for some reason: https://nodejs.org/api/esm.html#resolver-algorithm-specification (see `PACKAGE_RESOLVE` section there)

ukstv:
So, maybe change something on Jest side?

PaulLeCam:
Yes seems to work with the latest alpha so that's probably unrelated 👍 
I'll close it then.

ukstv:
Fun fact. That alpha release stopped working after rpc-utils update. We had to downgrade the dependency. Digging now into what is wrong about it.

PR #68, opened 2022-02-25
Add webpack to the CI pipeline to ensure browser support

PaulLeCam:
## Description

Mostly meant to avoid regressions affecting consuming packages when built-in node modules are used in dependencies.

## How Has This Been Tested?

Updated CI pipeline.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,3 +33,6 @@ jobs:
         run: npm run build
         env:
           CI: true
+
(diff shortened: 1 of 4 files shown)


ukstv:
To me it seems, like a good first step and/or cludge, in the absence of other tools. I have one idea, which I would like to work on this weekend. If after the coming weekend, I got no solution and/or you consider it inappropriate, we will merge this. Are you ok with it @PaulLeCam ?

PR #70, opened 2022-03-09
feat: add revocationPhaseOut option to verifyJWS

stephhuynh18:
When does https://ceramicnetwork.github.io/js-did/interfaces/verifyjwsoptions.html get updated? 

## How Has This Been Tested?
Added tests to `jws-verification-behavior.test`

### DIFF
--- a/src/__tests__/jws-verification-behavior.test.ts
+++ b/src/__tests__/jws-verification-behavior.test.ts
@@ -195,6 +195,9 @@ describe('atTime', () => {
 
   const beforeRotation = new Date('2021-07-07T08:00:19Z').valueOf()
   const afterRotation = new Date('2021-07-07T08:40:19Z').valueOf()
+  const timeKeysRotatedInSec = Math.floor(
+    new Date(VERSION_0_ROTATED.didDocumentMetadata.nextUpdate).valueOf() / 1000
+  )
 
   test('ok before rotation', async () => {
     const { kid } = await did.verifyJWS(jwsV0, { atTime: beforeRotation })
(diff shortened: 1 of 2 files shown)


src/__tests__/jws-verification-behavior.test.ts
```diff
@@ -210,6 +210,21 @@ describe('atTime', () => {
       did.verifyJWS(jwsV0, { atTime: afterRotation, disableTimecheck: true })
     ).resolves.toBeTruthy()
   })
+  test('ok after rotation if within revocationPhaseOut period', async () => {
+    const revocationPhaseOut =
+      (afterRotation - new Date(VERSION_0_ROTATED.didDocumentMetadata.nextUpdate).valueOf() + 1) /
+      1000
```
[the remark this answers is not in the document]
oed: Would maybe be clearer to maybe do something like this?
```js
const afterRotation = new Date(VERSION_0_ROTATED.didDocumentMetadata.nextUpdate).valueOf() + 10
const revocationPhaseOut = 11
```

oed (commented):
lgtm, testcases could be made a bit more clearer to read.

oed (approved):
Looks great. Thanks for the update!

src/did.ts
```diff
@@ -59,6 +59,11 @@ export interface VerifyJWSOptions {
    * Cacao OCAP to verify the JWS with.
    */
   capability?: Cacao
+
+  /**
+   * Number of seconds that a revoked key stays valid for after it was revoked
+   */
+  revocationPhaseOut?: number
```
stbrody, line 66: I have a strong belief that all time-based arguments should have their units included in the name.  Otherwise it's too easy to pass in milliseconds where seconds are expected, etc.  Can we rename this arg to `revocationPhaseOutSecs`?

stephhuynh18: I am okay with that but we do have: https://linear.app/3boxlabs/issue/NET-1184/use-proper-unix-timestamp-format-in-js-did

The renaming can/should be done there

stbrody: a perfect example of how easy it is to use the wrong units when they're not included in the argument names!

I'll file some tickets

PR #74, opened 2022-04-06
[NET-1304] Security implications and documentation around key revocation grace period

haardikk21:
## Description

Provides a first draft for documentation and some security implications around having a grace period for key revocation. 

## How Has This Been Tested?

N/A

## Definition of Done

- [ ] The work addresses the description and outcomes in the issue
- [ ] I have added relevant tests for new or updated functionality
- [ ] My code follows conventions, is well commented, and easy to understand
- [ ] My code builds and tests pass without any errors or warnings
- [ ] I have tagged the relevant reviewers
- [ ] I have updated the READMEs of affected packages
- [ ] I have made corresponding changes to the documentation
- [ ] The changes have been communicated to interested parties

## References:

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,17 +137,17 @@ const didWithCap2 = new DID({provider, resolver: KeyResolver.getResolver(), capa
 
 ## Security Considerations
 
-Ceramic allows for keys attached to DIDs to continue making updates to streams after revocation for a certain grace period. This is done to ensure that if multiple updates were committed to different nodes before the key was revoked, but the nodes chose to anchor commits at different times, the updates still make it through.
+Ceramic allows for keys attached to DIDs to continue making updates to streams after revocation for a certain grace period. This is done to avoid incorrectly rejecting valid signatures as being performed with a revoked key when in fact the key was valid at the time of the signature. This can happen due to the implementation details of how Ceramic determines when a DID issues a signature as well as how Ceramic determines when an authentication key for a DID is revoked. Since we cannot trust the system clock time on the client's machine when performing a signature or key revocation, Ceramic periodically "anchors" updates on a blockchain to get a timestamp before which we know the write must have happened. Consider the following scenario:
 
 1. Commit for an update is made to Node A
-2. Key revocation commit is made to Node B
+2. Node B's key revocation commit is anchored before node A's update
 3. Node A later tries to anchor the commit made earlier
 
 In this case, an external outlook might lead you to believe Node A's commit is invalid because the key had been revoked, but that's not necessarily true as in this case the commit was made earlier, just not anchored at that time. For this reason, a grace period is provided for updates created by revoked keys to still be verified successfully.
 
-Specifically, the `verifyJWS` function in `dids` uses `options.revocationPhaseOutSecs` as the seconds representing the grace period within which signatures authored by a given key will still be considered valid.
+Specifically, the `verifyJWS` function in `dids` uses `options.revocationPhaseOutSecs` as the seconds representing the grace period within which signatures authored by a given key that otherwise appears to be revoked will still be considered valid.
 
-While this approach has it's use cases, it also implies that an attacker could potentially author signatures from a stolen key and commit updates to streams using that stolen key. Even if the original owner revoked it, the attacker has a grace period within which they can make fake commits.
+While this approach helps solve the problem of incorrectly rejecting valid updates, it also implies that an attacker could potentially author signatures from a stolen key and commit updates to streams using that stolen key. Even if the original owner revoked the compromised key, the attacker has a grace period within which they can continue to make new commits. The best practice to mitigate this is to rotate keys associated to DIDs on a regular basis to make it less likely to have a compromised key associated with the DID.
 
 The `revocationPhaseOutSecs` value, therefore, should be set with this consideration in mind to a reasonable value. It either allows for a better UX for the owner and allows for a small window of attack for an attacker, or it does not allow for an attack window but can potentially mark certain commits as invalid due to latency in commits occuring.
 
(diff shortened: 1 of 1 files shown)


ukstv (approved):

README.md
```diff
@@ -135,6 +135,22 @@ const didWithCap = did.withCapability(cacao)
 const didWithCap2 = new DID({provider, resolver: KeyResolver.getResolver(), capability: cacao})
 ```
 
+## Security Considerations
+
+Ceramic allows for keys attached to DIDs to continue making updates to streams after revocation for a certain grace period. This is done to ensure that if multiple updates were committed to different nodes before the key was revoked, but the nodes chose to anchor commits at different times, the updates still make it through.
+
+1. Commit for an update is made to Node A
+2. Key revocation commit is made to Node B
+3. Node A later tries to anchor the commit made earlier
```
stbrody, line 144:
```suggestion
3. Node B's key revocation commit is anchored before node A's update
```

README.md
```diff
@@ -135,6 +135,22 @@ const didWithCap = did.withCapability(cacao)
 const didWithCap2 = new DID({provider, resolver: KeyResolver.getResolver(), capability: cacao})
 ```
 
+## Security Considerations
+
+Ceramic allows for keys attached to DIDs to continue making updates to streams after revocation for a certain grace period. This is done to ensure that if multiple updates were committed to different nodes before the key was revoked, but the nodes chose to anchor commits at different times, the updates still make it through.
```
stbrody, line 140: huh, this just made me realize.  Did we implement the key revocation grace period in a way that applies to all DIDs, not just 3IDs?  That seems like a mistake, the issues with key revocations and anchor timings are pretty specific to 3IDs.  @oed

stbrody, line 140:
```suggestion
Ceramic allows for keys attached to DIDs to continue making updates to streams after revocation for a certain grace period. This is done to avoid incorrectly rejecting valid signatures as being performed with a revoked key when in fact the key was valid at the time of the signature.  This can happen due to the implementation details of how Ceramic determines when a DID issued a signature as well as how the 3ID DID method built on Ceramic determines when an authentication key for a 3ID is revoked.  Since we cannot trust the system clock time on the client's machine when performing a signature or key revocation, Ceramic periodically "anchors" updates on a blockchain to get a timestamp before which we know the write must have happened.  Consider the following scenario:
```

README.md
```diff
@@ -135,6 +135,22 @@ const didWithCap = did.withCapability(cacao)
...
+
+1. Commit for an update is made to Node A
+2. Key revocation commit is made to Node B
+3. Node A later tries to anchor the commit made earlier
+
+In this case, an external outlook might lead you to believe Node A's commit is invalid because the key had been revoked, but that's not necessarily true as in this case the commit was made earlier, just not anchored at that time. For this reason, a grace period is provided for updates created by revoked keys to still be verified successfully.
```
stbrody, line 146:
```suggestion
In this case, an external outlook might lead you to believe Node A's commit is invalid because the key had been revoked when it was made, but that's not actually true as in this case the commit was in fact made earlier in real world time, it just wasn't assigned a timestamp by the anchoring system until after. For this reason, a grace period is provided for updates created by revoked keys to still be verified successfully.
```

README.md
```diff
@@ -135,6 +135,22 @@ const didWithCap = did.withCapability(cacao)
...
+
+Specifically, the `verifyJWS` function in `dids` uses `options.revocationPhaseOutSecs` as the seconds representing the grace period within which signatures authored by a given key will still be considered valid.
```
stbrody, line 148:
```suggestion
Specifically, the `verifyJWS` function in `dids` uses `options.revocationPhaseOutSecs` as the seconds representing the grace period within which signatures authored by a given key that otherwise appears to be revoked will still be considered valid.
```

README.md
```diff
@@ -135,6 +135,22 @@ const didWithCap = did.withCapability(cacao)
...
+
+While this approach has it's use cases, it also implies that an attacker could potentially author signatures from a stolen key and commit updates to streams using that stolen key. Even if the original owner revoked it, the attacker has a grace period within which they can make fake commits.
```
stbrody, line 150:
```suggestion
While this approach has it's use cases, it also implies that an attacker could potentially author signatures from a stolen key and commit updates to streams using that stolen key. Even if the original owner revoked it, the attacker has a grace period within which they can make fake commits.
```
```suggestion
While this approach helps solve the problem of incorrectly rejecting valid updates, it also implies that an attacker could potentially author signatures from a stolen key and commit updates to streams using that stolen key. Even if the original owner revoked the compromised key, the attacker has a grace period within which they can continue to make new commits.
```

stbrody (changes requested):
proposed some rewordings to add clarity and context for those who may not already be familiar with Ceramic and 3IDs when reading this README.

I'd like @oed to review this PR too before it is ultimately merged

README.md
haardikk21: thank you for these suggestions, very helpful. i need to get better at writing these kind of documents.

oed: I would say it applies to all DIDs since things signed right before the revocation might not be anchored yet. This is also relevant for CACAO, which I just realized doesn't use the revocation timeout yet:  https://github.com/ceramicnetwork/js-did/blob/main/src/did.ts#L334

oed: I don't think we need to explicitly mention 3ID here.

README.md
oed: Worth noting here is that best practice is likely to rotate keys on a regular basis to mitigate stolen keys.

oed (commented):
Looks mostly good. Left some comments. Feel free to merge when everything is addressed. 
Also added a new tech debt issue: https://linear.app/3boxlabs/issue/NET-1400/use-phase-out-for-cacao-verification

PR #75, opened 2022-04-11
chore: Update packages

ukstv:
Update versions of the dependencies

Apparently, we do not use `query-string` package anymore. And `ethers` is only dev dependency. Adjusted `package.json` accordingly.


PaulLeCam (approved):
Nice, thanks!

PR #77, opened 2022-04-26
feat: add parent did option/getter, for improved cacao support in con…

zachferland:
Reference [SIWE+ spec](https://www.notion.so/threebox/CACAO-DID-PKH-SIWE-Client-Libraries-5ce1f1f6e8c7417683c2b63e07f91cef#741885cca9d94532afee11a46621ab7d), allows ceramic to set stream controller to parent, while signing with the did instance. Allows devs to not have to know that implementation detail of cacao and allows glaze/apps libs to use cacao + other dids instances interchangeably.

### DIFF
--- a/src/did.ts
+++ b/src/did.ts
@@ -96,6 +96,7 @@ export interface DIDOptions {
   resolver?: Resolver | ResolverRegistry
   resolverOptions?: ResolverOptions
   capability?: Cacao
+  parent?: string
 }
 
 function isResolver(resolver: Resolver | ResolverRegistry): resolver is Resolver {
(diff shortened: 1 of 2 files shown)


src/did.ts
```diff
@@ -131,6 +139,23 @@ export class DID {
...
 
+  /**
+   * Get parent DID, parent DID is the capability issuer
+   */
+  get parent(): string {
+    if (!this._parentId) {
+      throw new Error('DID has no parent DID')
+    }
+    return this._parentId
+  }
```
oed, line 150: why is this not just a getter that might return undefined/null if `this._parentId` is not set?
Then we wouldn't need the `hasParent` getter either.

Also, if `this._capability` is set, the issuer of that should be returned right?

zachferland: yeah just pattern matching the other getters in here (did - id, authenticated, capability), but actually does seem better for types

and yes, parent set in constructor when given capability, and if given both they will match

ukstv (approved):
🏇

PR #80, opened 2022-05-31
feat: add cacao phaseout option support

zachferland:
passes option, tested in cacao package 

ref - https://github.com/ceramicnetwork/cacao/pull/10


ukstv (approved):

ukstv:
@zachferland Released as 3.2.0

PR #81, opened 2022-06-30
feat: throw on expired cacao on creation, not only validation

zachferland:
createJWS and createDagJWS throw when creating with expired cacao now 
could add other validation after, only for expiration for now

### DIFF
--- a/src/did.ts
+++ b/src/did.ts
@@ -304,10 +304,6 @@ export class DID {
     jws.link = compatibleCID
 
     if (this._capability) {
-      const exp = this._capability.p.exp
-      if (exp && Date.parse(exp) < Date.now()) {
-        throw new Error('Capability is expired, cannot create a valid signature')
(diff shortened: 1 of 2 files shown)


src/did.ts
```diff
@@ -300,6 +304,10 @@ export class DID {
     jws.link = compatibleCID
 
     if (this._capability) {
+      const exp = this._capability.p.exp
+      if (exp && Date.parse(exp) < Date.now()) {
+        throw new Error('Capability is expired, cannot create a valid signature')
+      }
```
oed, line 310: I don't think this is needed since this error would be thrown by `this.createJWS` which is called on line 296.

oed (approved):
lgtm, mod one comment

src/did.ts
zachferland: yup! removing, then merging

PR #82, opened 2022-07-06
build: initial mono repo re org, add did-session

zachferland:
Changes to monorepo, adds did-session

Tooling/builds align with what is in glaze, ie change to yarn, remove webpack, small config changes - @PaulLeCam  if i missed anything, or anything will change in our new repos, would like to align with those 

Biggest thing i was unsure about here, was the docs pipeline (downstream)? this switches them from html to markdown cc @ukstv @oed 

Also this uses a temporary glaze jest env from the glaze repo, would like to pull test dependencies from somewhere else and/or maybe move more integration tests to somewhere else (in a later pr). 

@glazed/did-session will become did-session, and few breaking changes will be made in later PR before it is released under that package name.


PaulLeCam (approved):
LGTM from a quick look, one thing about the doc generation, in order to have the `dids` docs display `dids/src`, you can set `@module dids` at the end of the header doc comments in `dids/src/index.ts`, similar to how it's done for `did-session`.
About the docs setup I think we need a wider discussion about how to manage docs across our repos in a more unified way rather than making ad-hoc decisions for each repo.

PaulLeCam (approved):

zachferland:
thanks @PaulLeCam, changed that, and docs specs today made sense, going to merge in a bit

PR #83, opened 2022-07-12
Feat/cleanup session api

zachferland:


zachferland:
partly follow up to https://github.com/ceramicstudio/js-glaze/pull/134

zachferland:
cc @pawartur for api changes mentioned before

packages/did-session/src/index.ts
```diff
@@ -71,52 +69,108 @@
...
+ * the domain name of your app. If you are using the library in a non web based environment you will need to pass
+ * the `domain` option otherwise an error will thrown.
+ *
+ * ```ts
+ * const session = await DIDSession.authorize(authProvider, { resources: [...], domain: 'YourAppName' })
+ * ```
+ *
+ * ## Typical usage pattern
  *
  * A typical pattern is to store a serialized session in local storage and load on use if available. Then
```
PaulLeCam, line 107: Could we add some warnings about how session management is the responsibility of the app and they should make sure to understand the security implications of storing sessions in browsers please?
Maybe we can mention that local storage is used here for illustrative purposes but not a recommended mechanism?

PaulLeCam (approved):
Looks great, the API changes are very nice!
Just wrote a comment about session management, I assume most apps will just use local storage but could we add some warnings about it please?

packages/did-session/src/index.ts
zachferland: yeah good idea, more info on what is stored/risks is useful although idk of other implementation options (at least now) 

will add section before merging

PaulLeCam: Thanks!

PR #89, opened 2022-08-09
Update repository setup

PaulLeCam:
Updates the setup to be the same as ComposeDB, notably:
- Uses pnpm instead of Yarn, which helps identify missing dependencies issues from hoisted dependencies
- TypeDoc + Docusaurus setup for docs, just with basic configuration for the API docs, pages will need to be edited

Related to the docs generation setup, I changed `interface`s in the `dids` package to be `type` definition instead, as TypeDoc creates dedicated pages for interfaces but puts the types in the module page, that allows to reduce the number of pages generated.

I also kept the `docs` folder with the markdown pages for now until the website is fully setup and deployed.


zachferland (approved):
Looks good, thanks for aligning, whats does it mean "pages will need to be edited" for Docusaurus? just mean content besides api/reference output docs?

PaulLeCam:
Yes I kept the default pages from Docusaurus so they'll need to be removed and replaced by the pages we want, and same for the overall config it's mostly using the default names and links to Docusaurus rather than the js-did repo for example.
Essentially these content pages could be used to present a simple getting started with installing and configuring the packages, maybe some guides about the differences between the different DID methods, how to use DID session, etc?

zachferland:
ok, yup! makes sense

PR #104, opened 2022-09-07
Feat/move package cacao

zachferland:
Moves `ceramic-cacao` and `key-did-provider-ed25519` to this monorepo. 

The remaining libraries `key-did-resolver` and `pkh-did-resolver` required more type/code changes to use lint/tool configuration here, so decided to remove them for now from this pr and prioritize later again. cacao was most important to move for now. Additionally `key-did-provider-secp256k1` was also removed for now, similarly that lib requires more changes as it has not been updated in a while.


PaulLeCam (approved):
LGTM, just 2 related things than can be changed in this PR or follow-up ones I think:
- If possible, let's use [pnpm's `workspace` protocol](https://pnpm.io/workspaces#workspace-protocol-workspace) to define dependencies between packages in the monorepo, that ensures these packages are in sync rather than downloading from npm if there's a version mismatch.
- Can we add API docs and possibly guides to the docs website for these added packages please?

pawartur:
I can help adding the docs, @zachferland :)

zachferland:
thanks @PaulLeCam @pawartur 

yeah definitely want to use those pnpm workspaces, will add in next pr 

will plan to archive the key-did-provider-ed25519 and cacao repos soon

PR #106, opened 2022-09-19
Feat/authmethods

zachferland:
Implement authmethods and verifiers as defined in https://www.notion.so/threebox/CACAO-Auth-Verify-3a5b855f72684da78d707e44503786f8. 

Two things changed from spec: 
1) authmethod/verifiers in their own libraries, as discussed in comments 
2) authmethod consumes siwx args/opts instead of cacao opts, mostly for readability (should always be able to map one to the other) 

This is a breaking change for did-session again, but we could make this the 1.0.0 release now. Alternatively we could have did-session consume both authproviders and authmethods, but dont see the need for that. 

The solana authmethod will come in another pr

Future Improvements: 
[] split tests out across relevant libraries and increase coverage, right now most tests from cacao were just moved to "integration" tests, so that we could have the same tests from before pass first 
[] need to consider naming before release, a change for `ceramic-cacaos`, and naming for blockchain packages
[] some libs moved from less strict lint/type env could be improved


docs/modules/dids.md
```diff
@@ -8,6 +10,145 @@ A simple library to interact with DIDs that conform to the DID-provider interfac
 npm install dids
 ```
 
+## Usage
+
+### Authentication with the provider
+
+```js
+import { DID } from 'dids'
+import Ed25519Provider from 'key-did-provider-ed25519'
```
oed, line 19: I think this import doesn't have a default
```suggestion
import { Ed25519Provider } from 'key-did-provider-ed25519'
```

PaulLeCam (changes requested):
Could we use a npm namespaces for the CACAO packages please?
I just registered the `@cacao` scope on npm and added you @zachferland, maybe we could use it?

packages/ethereum/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+```
+npm install --save ethereum-cacao
+```
+
+## Usage
+
+```ts
+// Auth Usage
+import { EthereumWebAuth, getAccountId } from 'ethereum-cacao'
```
oed, line 14:
```suggestion
import { EthereumWebAuth, getAccountId } from 'did-pkh-eip155'
```

Rather than calling these packages something related to CACAO I think using DID PKH is more appropriate.

packages/solana/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "solana-cacao",
```
oed, line 2:
```suggestion
  "name": "did-pkh-solana",
```

oed (commented):
Open to better suggestions around naming, but I don't think CACAO is the right thing here. 
Ideally we would have a `did` npm org, but looks like it's taken.

PaulLeCam:
Yes `@did`, `@dids` and similar are taken, but I registered `@identifiers` and `@identities` we could use as alternatives potentially?

oed:
How about `@didtools`? Basically wondering if there's a namespace we can use for most pkgs in this repo?
Also what's the benefit of putting all packages within a namespace? As long as they are within the same npm org you should be able to easily manage publish access?

PaulLeCam:
I like `@didtools`, and yes would be good to use the same namespace for most packages if possible!

The main advantages of using the same namespace IMO are on the authoring side, being able to use consistent naming conventions independently of names already taken on npm, and on the consuming side having more "guaranties" that the packages you want to use are the expected ones rather than possible mistakes (for example, using `cacao-ceramic` instead of `ceramic-cacao`).

packages/cacao/src/cacao.ts
```diff
@@ -1,26 +1,49 @@
...
+
+export type SignedCacao = {
+  h: Header
+  p: Payload
+  s: Signature
+}
+
+export type Verifiers = Record<string, CACAOVerifier>
+
+export type CACAOVerifier = (cacao: Cacao, opts: VerifyOptions) => Promise<void>
```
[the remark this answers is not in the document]
zachferland: yeah good idea

docs/modules/dids.md
```diff
@@ -8,6 +10,145 @@ A simple library to interact with DIDs that conform to the DID-provider interfac
...
+// See https://github.com/decentralized-identity/did-resolver
+const did = new DID({ resolver: KeyResolver.getResolver() })
+
+// Resolve a DID document
+await did.resolve('did:key:...')
+```
+
+### Creating a DID with attached CACAO
+
+Using CACAO OCAPs to create a Key DID that can sign on behalf of a PKH DID.
```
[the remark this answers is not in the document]
zachferland: yeah someone would have to create a general did authmethod/verifier first though, but yes can start to move to less specific references

zachferland:
thanks @PaulLeCam @oed 

think its reasonable to keep current names for dids libraries (but open to namespacing), mostly want some standard naming for authmethods and to change cacao lib name.

could see @cacao for those, as mentioned above
`@cacao/core`
`@cacao/ethereum`
`@cacao/solana`

or 
`@didpkh/ethereum`
`@didpkh/solana`

packages/ethereum/README.md
zachferland: yeah kinda torn on this, i think did pkh wording is more useful to relay its purpose to developers, but technically its very specific to cacao, you could have more general did pkh signers/verifier after

docs/modules/dids.md
```diff
@@ -8,6 +10,145 @@ A simple library to interact with DIDs that conform to the DID-provider interfac
...
+import KeyResolver from 'key-did-resolver'
+import { Cacao, SiweMessage } from 'ceramic-cacao'
+
+const seed = // 32 bytes of entropy, Uint8Array
+const provider = new Ed25519Provider(seed)
+
+const siwe = new SiweMessage("...");
+const cacao = Cacao.fromSiweMessage(siwe);
+
+// Add capability to existing DID instance
```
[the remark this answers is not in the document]
zachferland: not mutated, just a static helper that returns an new did instance with the cap now, could consider removing at another point (no other breaking changes in dids here right now)

oed:
I don't think `@cacao/*` makes much sense. How about this:

`@didtools/pkh-did-eip155`
`@didtools/pkh-did-solana`
`@didtools/cacao`

packages/ethereum/README.md
oed: Imo, catering to devs should be top prio.

PaulLeCam:
Maybe we could remove `did` from these package names, as it's already part of the namespace? Ex: `@didtools/pkh-did-solana` -> `@didtools/pkh-solana`, or `@didtools/pkh-method-solana` to be more explicit?

Regarding the other package names, I think we could keep `dids` as the main one, but maybe we could rename the other ones eventually? Ex: `key-did-resolver` -> `@didtools/key-resolver`?
I'm also split on `did-session`, we already changed its name once but now that we're planning on a v1 with breaking changes, maybe we can also use the same convention and rename it to `@didtools/session`?

oed:
Yep, @PaulLeCam I updated the comment above with modified names.

zachferland:
yeah but these libraries are not general did pkh libraries, they are specifically for signing/verifying a cacao, not any signed payload. But agree with these names anyways, think pkh naming is more useful/info for devs, plus only usable way to use did pkhs right now, and could be more general did pkh libraries at a later point.

Going to change names as suggested in joels comment, and keep others for now. Can still consider changing did-session before release, but lean towards keeping for now.

zachferland:
Updated package names

PaulLeCam (approved):
Thanks for the changes!

oed:
Did we register the `@didtools` namespace?

PaulLeCam:
> Did we register the `@didtools` namespace?

I just did and added both of you and Artur, you should get invites from npm.

oed:
@PaulLeCam I guess add Zach as well?

ukstv (approved):

PaulLeCam:
> @PaulLeCam I guess add Zach as well?

Yes I added him already!

zachferland:
merging, dids and all @didtools/* have been released, did-session will be released after docs updates

PR #113, opened 2022-09-27
docs: up docs website for did-session v1

zachferland:


oed (approved):
lgtm, but check is failing.

PR #117, opened 2022-10-05
Docs/add more sol auth

zachferland:
Improves pkh-ethereum docs
Add pkh-solana docs 
Adds to Readme and docs website output


zachferland:
Merging so i can share some of the docs, tagged reviews if anyone wants to look still

PR #123, opened 2022-11-02
Fixed typos in pkh_ethereum docs

pawartur:
Fixed typos in pkh_ethereum docs around getting ethereum addresses and auth method for didsession

### DIFF
--- a/docs/modules/did_session.md
+++ b/docs/modules/did_session.md
@@ -23,7 +23,7 @@ import { DIDSession } from 'did-session'
 import { EthereumWebAuth, getAccountId } from '@didtools/pkh-ethereum'
 
 const ethProvider = // import/get your web3 eth provider
-const addresses = await ethProvider.enable()
+const addresses = await (ethProvider as any).request({ method: 'eth_requestAccounts' })
 const accountId = await getAccountId(ethProvider, addresses[0])
 const authMethod = await EthereumWebAuth.getAuthMethod(ethprovider, accountId)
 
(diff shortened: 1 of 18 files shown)


docs/modules/did_session.md
```diff
@@ -23,7 +23,7 @@ import { DIDSession } from 'did-session'
 import { EthereumWebAuth, getAccountId } from '@didtools/pkh-ethereum'
 
 const ethProvider = // import/get your web3 eth provider
-const addresses = await ethProvider.enable()
+const addresses = await (ethProvider as any).request({ method: 'eth_requestAccounts' })
```
oed, line 26: Don't think the example is using typescript?

pawartur: Right, sorry.

zachferland: all the snippets are marked as ts in markdown, but wasnt very intentionally chosen, mostly just because types were sometimes referenced but mostly the snippets are just plain js

zachferland:
thanks @pawartur! this is related to `.enable` being deprecated right?

docs/modules/did_session.md
pawartur: I can convert all the snippets to `js` then :). Did I understand correctly that this is what we want?

pawartur:
> thanks @pawartur! this is related to `.enable` being deprecated right?

Yes, I think so. I did these changes, because Joel asked me to replace `.enable()`, plus I saw that we were missing `await`s in a few places :)

docs/modules/did_session.md
zachferland: yeah think that makes sense here

zachferland (approved):

PR #125, opened 2022-11-02
feat: add key did provider secp256k1 package

zachferland:
Moves https://github.com/ceramicnetwork/key-did-provider-secp256k1, but that package was a bit out of date, this mostly copies over everything from `key-did-provider-ed25519` for `secp256k1` instead


oed (approved):
lgtm, Are we making a new release after merge?

zachferland:
yeah we can release, I dont have permissions for this package if you wan to add them, cc @oed

PR #127, opened 2022-11-04
add tezos support

lhd791150351:
# Add Tezos support

## Description

- added a `cacao/src/siwx/siwTezos.ts` file and extend `SiwxMessage`
- implemented `fromSiwTezosMessage` function in `src/cacao.ts`
- added a `pkh-tezos` repo that contains `Authmethods` and `Verifier`


packages/pkh-tezos/src/verifier.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  const publicKey = await publicKeyFinder(issuerAddress)
+
+  if (!verifySignature(payload, publicKey, signature)) {
+    throw new Error(`Signature does not belong to issuer`)
+  }
+}
+
+async function publicKeyFinder(address: string): Promise<string> {
+  const response = await fetch(`https://api.tzstats.com/explorer/account/${address}`)
```
oed, line 39: Is there no way to verify signatures without relying on this external api?

oed (commented):
Super excited about Tezos support!

Before this can be merged Tezos need to be supported in the chain agnostic namespeces. In particular **10** and **122**: https://namespaces.chainagnostic.org/

packages/pkh-tezos/src/verifier.ts
lhd791150351: I'm sorry, we haven't found any other solutions yet.

oed: It might be that there isn't a good solution to this unfortunately.

packages/pkh-tezos/src/authmethod.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+async function createCACAO(
+  opts: AuthMethodOpts,
+  tzProvider: any,
+  account: AccountId,
+  publicKey: string
+): Promise<Cacao> {
+  const now = new Date()
+  const oneDayLater = new Date(now.getTime() + 24 * 60 * 60 * 1000)
```
zachferland, line 65: we changed to 1 week in the other libraries now, if you can do the same here 
```suggestion
  const now = new Date()
  const oneWeekLater = new Date(now.getTime() + & * 24 * 60 * 60 * 1000)
```

packages/pkh-tezos/src/authmethod.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+type TezosNetwork = 'mainnet' | 'devnet'
+
+export namespace TezosWebAuth {
+  // eslint-disable-next-line @typescript-eslint/require-await
+  export async function getAuthMethod(
+    tzProvider: any,
+    account: AccountId,
+    publicKey: string
```
zachferland, line 22: earlier comment mentioned pk lookup and network request, the approach now is the dev passes the pubkey here right? and they can look it up how they want 

this approach looks good

packages/pkh-tezos/src/authmethod.ts
```diff
@@ -0,0 +1,99 @@
...
+  const signature = await sign(tzProvider, signData)
+  siwTezosMessage.signature = signature + publicKey
+  return Cacao.fromSiwTezosMessage(siwTezosMessage)
+}
+
+export function requestChainId(network: TezosNetwork): string {
+  return chainIdMap[network]
+}
+
+export function getAccountId(network: TezosNetwork, address: string): AccountId {
```
zachferland, line 90: these account helper functions are optional and you can add what ever you want to make configuring the authmethod easier, this is the place you can add network requests if you want

For example in solana `getAccountIdByNetwork` is same as here, but `getAccountId` does a network lookup using the provider, you could take the same approach here and you could add a helper for get the pubkey from address here (we just want to avoid/limit network requests in the verifiers) 

Summarize:
- getAccountIdByNetwork keep same
- getAccountId take a tezos provider and make network request to get chain id 
- add getPublicKey helper, and similarly take tezos provider and make network request to get chain id

zachferland (commented):
Overall looks good! excited to get this merged, had some minor comments/changes 

Since we really havent added tests in these packages yet, how did you go about using/testing this so far? and how does the signature request look in a tezos wallet?

packages/pkh-tezos/src/authmethod.ts
QibingLee: Yes. Now we have added the getPublicKey helper, directly using tzProvider to get public key

packages/pkh-tezos/src/verifier.ts
QibingLee: Solved!

packages/pkh-tezos/src/authmethod.ts
QibingLee: Fixed.

packages/pkh-tezos/src/authmethod.ts
QibingLee: Fixed.

QibingLee:
Hi, @zachferland I sent you some figures in Discord, including signature request and console results. You can also make a test using Template Wallet and run the [demo](https://github.com/lhd791150351/tezos-ceramic/blob/master/demo/src/App.tsx). Note that we only support tz1 account for now (ed25519), because it is enough for web users. Template uses tz1 as the default account, and it is the best Chrome wallet now (I cannot find a second usable version).

zachferland:
Thanks for the the changes, helper methods look good now and docs and everything else does as well. Think its good with just tz1 if that is mostly typically used. I might spend some time to try it out, but should be good to merge/release beginning of next week.

zachferland (approved):

QibingLee:
Hi, @zachferland. For the Template wallet, it did not recognize the '\n', so the first look of SIWE message is not that formatted. I think it depends on the wallet, maybe another works.

zachferland:
Yeah partly up to wallets and standards in ecosystem on how to display messages like this, think it makes sense to keep this way (with new line breaks), other ecosystem wallets support it

Merging, and will release a first version

Thanks again for adding tezos support! 🚀🎉

zachferland:
Actually, a few build issues, will see if i can fix quickly first

zachferland:
Just lock file, going to merge and update lock file on main

PR #128, opened 2022-11-10
build: change package name, secp25k1 provider

zachferland:
Changes package name to @didtools/key-secp256k1 and released.


oed (approved):
lgtm

PR #130, opened 2022-11-21
feat: increase default cacao expiry to one week (from 1 day)

zachferland:
This PR increases the default cacao expiry when used with authmethods in did-session from 1 day to 7 days. We still have the 1 day revocation phaseout, effectively making it 8 days. Expiry is just a client default, not in the cacao library or enforced in anyways, clients may already configure their own but expect most rely on default. Default phaseout is more likely be recognized, as few would change that config value. 

We can discuss the time interval we want to use here. Also how we want to version the release.

### DIFF
--- a/packages/pkh-ethereum/src/authmethod.ts
+++ b/packages/pkh-ethereum/src/authmethod.ts
@@ -54,7 +54,7 @@ async function createCACAO(
   account: AccountId
 ): Promise<Cacao> {
   const now = new Date()
-  const oneDayLater = new Date(now.getTime() + 24 * 60 * 60 * 1000)
+  const oneWeekLater = new Date(now.getTime() + 7 * 24 * 60 * 60 * 1000)
 
   const siweMessage = new SiweMessage({
     domain: opts.domain,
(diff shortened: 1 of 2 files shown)


oed:
One weeks is a good default imo. 
Patch version should be enough to bump here?

stbrody:
I'm really conflicted about this.  1 week seems like a really long time to let an app hold onto credentials to write data on your behalf, especially while we have no way to revoke a granted credential.  But given the issues we've had with anchoring and CACAO expirations, I guess defaults with weaker security properties is better than data loss.

As long as we feel comfortable changing this default back to something lower in the future, once anchoring is more robust, then I guess this is fine for now.

CC @ukstv @3avi for visibility.

oed:
Fyi @stbrody 3IDv2 will most likely include a way to revoke capabilities!

oed (approved):

zachferland:
Going to merge and release tmrrw, seems 1 week is good, will release as patch, can change behavior/expectations for app, but prefer patch so more apps pull it in, and extending should limit any issues with app logic (vs decreasing)

PR #138, opened 2023-01-13
Add Sui Support

0xzoz:
# [Add Sui Support]

## Description

> This was created due to an initial PR created [here](https://github.com/ceramicnetwork/js-ceramic/pull/2638) and the library no longer being supported

This PR is to add SUI wallet support. [Sui](https://sui.io/) is a new layer 1 blockchain developed by Mysten Labs, it is in the early stages of launching and is part of the Move ecosystem that originated from the development work around [Diem](https://www.diem.com/en-us/) created at [Meta](https://investor.fb.com/home/default.aspx)

The PR adds the `pkh-sui` package and additional changes to the `cacao` package to support it.

It is being added here as a draft in case the team wants to provide context or comments. Some unit testing is done but it is awaiting local development testing using Sui wallets and guidance around the addition of CAIP's for Sui network(see below)

## How Has This Been Tested?

Local unit testing
- run local unit test

Describe the tests that you ran to verify your changes. Provide instructions for reproduction.

- [ X ] sui.test.ts in the respective packages (e.g. Add new tests to cover additional code)
- [ ] Local dev testing - TBD

## PR checklist

- [ ] I have tagged the relevant reviewers and interested parties - N/A
- [ ] I have updated the READMEs of affected packages - N/A
- [ ] I have made corresponding changes to the documentation - N/A

## References:

> This PR is dependent on the following CAIP's and will require guidance from the Sui team.

- [CAIP](https://github.com/ChainAgnostic/CAIPs/pull/190)
- [Relevant Sui discussion around CAIP](https://github.com/MystenLabs/sui/issues/6624)

### DIFF
--- a/packages/pkh-sui/src/authmethod.ts
+++ b/packages/pkh-sui/src/authmethod.ts
@@ -1,5 +1,6 @@
 import { AccountId } from 'caip'
 import { randomString } from '@stablelib/random'
+import { toString } from 'uint8arrays/to-string'
 import { Cacao, SiwSuiMessage, AuthMethod, AuthMethodOpts } from '@didtools/cacao'
 
 export const SUI_MAINNET_CHAIN_REF = 'mainnet' // TBD when CAIP-2 is finalized
@@ -37,7 +38,7 @@ export function assertSupportedProvider(suiProvider: any): asserts suiProvider i
 async function sign(suiProvider: any, message: Uint8Array) {
   assertSupportedProvider(suiProvider)
   const { signature } = await suiProvider.signMessage(message)
-  return signature
+  return toString(signature, 'base64')
 }
 
 async function createCACAO(
--- a/packages/pkh-sui/src/authmethod.ts
+++ b/packages/pkh-sui/src/authmethod.ts
@@ -1,0 +1,77 @@
+import { AccountId } from 'caip'
+import { randomString } from '@stablelib/random'
+import { Cacao, SiwSuiMessage, AuthMethod, AuthMethodOpts } from '@didtools/cacao'
+
+export const SUI_MAINNET_CHAIN_REF = 'mainnet' // TBD when CAIP-2 is finalized
+export const VERSION = '1'
+export const CHAIN_NAMESPACE = 'sui'
+
+export namespace SuiWebAuth {
+  // eslint-disable-next-line @typescript-eslint/require-await
+  export async function getAuthMethod(suiProvider: any, account: AccountId): Promise<AuthMethod> {
+    if (typeof window === 'undefined')
+      throw new Error('Web Auth method requires browser environment')
+    const domain = (window as Window).location.hostname
+
+    return async (opts: AuthMethodOpts): Promise<Cacao> => {
+      opts.domain = domain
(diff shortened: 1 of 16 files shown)


packages/pkh-sui/src/authmethod.ts
```diff
@@ -0,0 +1,97 @@
...
+    resources: opts.resources,
+  })
+
+  const signData = siwSuiMessage.signMessage()
+  const signature = await sign(suiProvider, signData)
+  siwSuiMessage.signature = signature
+  return Cacao.fromSiwSuiMessage(siwSuiMessage)
+}
+
+export async function getAccountId(suiProvider: any, address: string): Promise<AccountId> {
```
zachferland, line 79: If it is not possible to request the network from the provider, then you can remove this function, and just use `getAccountIdByNetwork`, as long as most of the networks are covered. All these functions are just optional helpers for each chain depending on implementation/needs

zachferland (commented):
Overall looks good so far, would mostly need to further define CAIP specs, have others review and reflect those changes/specs here.

packages/pkh-sui/src/verifier.ts
```diff
@@ -0,0 +1,40 @@
...
+  export function getSuiVerifier(): Verifiers {
+    return {
+      // eslint-disable-next-line @typescript-eslint/require-await
+      'sui:ed25519': async (cacao: Cacao, opts: VerifyOptions): Promise<void> => {
+        verifySuiSignature(cacao, opts)
+      },
+    }
+  }
+  
+  export function verifySuiSignature(cacao: Cacao, options: VerifyOptions) {
```
0xzoz, line 25: @zachferland Ive been looking into this more and it appears there is not currently a method to get the public key from the sui address to verify the signature. Im looking for insight from you to see what you would consider the best method for approaching this problem?

packages/pkh-sui/src/authmethod.ts
0xzoz: Yes, this is possible. Will change this when I get clarity on the `verifySuiSignature` method alternatives.

packages/pkh-sui/src/verifier.ts
zachferland: is secp256k1 not also supported? then ecrecover from signature can be used (like eth)

0xzoz: keys can be either ed25519, secp256k1, secp256r1

zachferland: do wallets support all three? is one default? how does that work typically

if these messages could just use secp256k1, that would be great

0xzoz: Wallets can be one of the 3 and it is specified at account creation. It is set by a flag when the a keypair is created programmatically.

ben-passion-xyz:
Hi, is there any progress with Sui support? I'm considering using composedb with Sui chain. I may not be familiar with signing algorithms, but is there any further information I can learn or help with?

Akagi201:
any update on this?

PR #139, opened 2023-01-13
feat(recaps): add proposed api

zachferland:
Opened as ongoing DRAFT: 

Currently just includes the proposed API for first iteration (to be implemented soon).

### DIFF
--- a/packages/recaps/src/index.ts
+++ b/packages/recaps/src/index.ts
@@ -1,3 +1,6 @@
+import { encodeBase64Url, base64urlToJSON } from 'did'
+import * as u8a from 'uint8arrays'
+
 type CeramicWrite = {
   "ceramic/write": []
 }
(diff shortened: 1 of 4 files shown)


packages/recaps/src/index.ts
```diff
@@ -0,0 +1,93 @@
+type CeramicWrite = {
+  "ceramic/write": []
+}
```
oed, line 3: Do we want to keep Ceramic specific things outside of here? Currently all Ceramic specific code of SIWx lives in the js-ceramic codebase.

oed (commented):
Seems reasonable.

packages/recaps/src/index.ts
```diff
@@ -0,0 +1,129 @@
...
+    validCeramicTarget(target)
+    return new ReCap(this._caps.concat([[target, action]]))
+  }
+
+  /**
+   * Verify that a ReCap Statement is valid for this ReCap and encodes the same delegations
+   */
+  // TODO att/caps may not be ordered, unless spec says otherwise
+  verifyStatment(statement:string, delegee: string):boolean {
+    return this.toStatement(delegee) === statement
```
oed, line 112: Note that the complete SIWE statement may contain additional data after the ReCap statement. Might make sense to do `startsWith` instead.
```suggestion
    return statement.startsWith(this.toStatement(delegee))
```

packages/recaps/src/index.ts
zachferland: Mostly likely in the end, but thinking that for now everything ceramic specific goes here for now (to keep more self contained while experimental), and more useful to drive with specific use case first (ceramic), then generalize after 

even if specs dont have namespaces, a general library will need to have some form of that, similar to spruce libs, lots of the rules/parsing are defined by that 

likely will make sense to combine most of siwx + recap things together instead this separate library here

packages/recaps/src/index.ts
zachferland: yeah thats a good point, this does assume only the recap statement and not the siwx statement. As mentioned above, think siwx + recap stuff will end up together, and this would validated like this at that level.

PR #142, opened 2023-02-07
fix: normalize checksum address eth

zachferland:
Checksum (mixed case) addresses in eth auth where no longer normalized in switch to did-session. The effect is more minimal here than for caip10 links, and less likely to come up, as it would require the same account to use to two different wallets/providers, which is unlikely (two returning diff format addresses, same wallets either case would not throw an error). It would/will be more problematic for general queries/relations. 

This adds normalization for eth address and is technically a breaking change, and existing checksum address providers that did work fine, will now break, no longer allow content to be added to those streams. It should be few to none returning this, but should try to capture before merging (PR in draft for now)

### DIFF
--- a/packages/pkh-ethereum/src/authmethod.ts
+++ b/packages/pkh-ethereum/src/authmethod.ts
@@ -1,7 +1,7 @@
 import { Cacao, SiweMessage, AuthMethod, AuthMethodOpts } from '@didtools/cacao'
 import { randomString } from '@stablelib/random'
 import { AccountId } from 'caip'
-import { safeSend } from './utils.js'
+import { safeSend, normalizeAccountId } from './utils.js'
 
 /**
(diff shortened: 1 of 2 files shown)


oed:
Yeah, Can we try to capture who would be affected by this and how?

zachferland:
yeah looking into it today

zachferland:
Going to capture metrics on it this week and then will follow up

zachferland:
with v2 release of dids, think this would be a good time to merge/release this, to pair with that, thoughts @oed ?

oed:
Yes, we should release this sooner rather than later

zachferland:
Removed from draft, will merge/release this afternoon 

effects should be fairly limited, and recent major version release will help further minimize

PR #147, opened 2023-02-24
feat(pkh-tezos): remove @taquito/utils for stablelib

zachferland:
need to test/review further

### DIFF
--- a/packages/pkh-tezos/test/index.test.ts
+++ b/packages/pkh-tezos/test/index.test.ts
@@ -1,5 +1,28 @@
-describe('Solana Cacao Auth Verify', () => {
-  test.skip('Verify', () => {
+import { InMemorySigner } from '@taquito/signer'
+import { verifySignature, getPkhfromPk } from '../src/verifier'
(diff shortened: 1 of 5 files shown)


zachferland:
Clearly not working, add simple tests for the two funcs i tried to implement 

maybe some differences between handling uin8arrays/enc/dec, nothing obviously wrong yet

packages/pkh-tezos/test/index.test.ts
```diff
@@ -1,5 +1,28 @@
-describe('Solana Cacao Auth Verify', () => {
-  test.skip('Verify', () => {
+import { InMemorySigner } from '@taquito/signer'
+import { verifySignature, getPkhfromPk } from '../src/verifier'
+import { char2Bytes, verifySignature as taquitoVerifySignature, getPkhfromPk as taquitoGetPkhfromPk } from '@taquito/utils'
```
oed, line 3: Can we just generate some test vectors so that we don't need to include the `@taquito` deps at all?

oed (changes requested):
Lgtm, just one request

oed (approved):
Lets actually merge this and get into js-ceramic asap. We can address test vectors later

Issue #131, opened 2022-12-12
Session DID Changes In Reauthentication Preventing Stream Updates


alex-mccreary:
# Description

If I authenticate with DIDSession.authorize using an EthereumAuthProvider, I am provided with a session.did.id and session.did.parent. However, when I re-authenticate later on, a different session.did.id is provided; the session.did.parent remains the same. This prevents me from updating any streams created by the previous session.did.id

# Technical Information
Versions
-"@ceramicnetwork/blockchain-utils-linking": "^2.3.0",
-"@ceramicnetwork/http-client": "^2.3.0",
-"did-session": "^0.1.1",

1. Create a DIDSession using an EthereumAuthProvider
2. Create a stream and set the controller as `controllers: [this.ceramic.did.id]`
3. Re-authenticate the DIDSession
4. Update the previously created stream and encounter an error:
`{"error":"Can not verify signature for commit bagcqcera4i776kf624dvn5h2avj37jjgepgbqd2yw3fdilpkttllwxfj42gq: invalid_jws: not a valid verificationMethod for issuer: did:key:z6MkmSwWYbLtgRVoAoVpaeHafXHWuGvDAt1nzMvRtxjMAfiT#z6MkmSwWYbLtgRVoAoVpaeHafXHWuGvDAt1nzMvRtxjMAfiT"`

I also tried setting the parent as the controller like:
`controllers: [this.ceramic.did.parent]`, but when creating a NEW stream, this resulted in:
`{"error":"Can not verify signature for commit bagcqcera7og333rcoajf4uvujlskpcxfjpqtnmknmbiu7ersrf42gvw32ftq: Signature does not belong to issuer"}`

It seems others on Discord were encountering this same issue:
![image](https://user-images.githubusercontent.com/109529190/207107931-cff61a94-5b59-4e22-be22-32f6b5a5aeb9.png)

zachferland:
Not sure about the last error yet, that should work from description, but this the expected behavior for did session as Mohsin described in the commend above. The did you are signing with is a temporary session key that will change with every session, the controller should be your did:pkh (blockchain account), which of course is always the same. 

could you try creating and updating a stream with did-session without changing the controller? ceramic/libs will handle the dids/controllers, no need to pass in that option, any reason you chose to pass that in? 

also could you update did-session to v1 and did-session no longer uses `blockchain-utils-linking`, docs are up to date - https://did.js.org/docs/api/modules/did_session

alex-mccreary:
[...]

PR #148, opened 2023-03-01
fix: eth personal sign hex message

zachferland:
personal_sign spec expects a hex encoded utf8 message, most providers do seem to support both hex & utf8 strings (ie metamask provider and we havent had any other reported errors from common providers), plus most libraries (ethers, web3) accept both and encode as hex before making provider call 

this will provide better general support, esp any other providers that may just follow spec 

This will resolve issue - https://github.com/ceramicnetwork/js-did/issues/131

### DIFF
--- a/packages/pkh-ethereum/package.json
+++ b/packages/pkh-ethereum/package.json
@@ -50,6 +50,7 @@
     "@didtools/cacao": "workspace:^2.0.0",
     "@ethersproject/wallet": "^5.7.0",
     "@stablelib/random": "^1.0.2",
-    "caip": "^1.1.0"
+    "caip": "^1.1.0",
+    "uint8arrays": "^4.0.3"
   }
 }
--- a/packages/pkh-ethereum/src/authmethod.ts
+++ b/packages/pkh-ethereum/src/authmethod.ts
@@ -2,6 +2,7 @@ import { Cacao, SiweMessage, AuthMethod, AuthMethodOpts } from '@didtools/cacao'
 import { randomString } from '@stablelib/random'
 import { AccountId } from 'caip'
 import { safeSend, normalizeAccountId } from './utils.js'
+import * as u8a from 'uint8arrays'
 
 /**
  * SIWX Version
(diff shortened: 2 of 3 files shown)


oed (commented):
How has this been tested?

zachferland:
@oed the existing integration tests use pkh-ethereum, they still pass and most importantly the verifiers have not changed in any way and still pass (will not be breaking, nor have to be coordinated with ceramic nodes in anyway) 

can pretty quickly try with metamask as well here to see that utf8 string is presented to user still - https://metamask.github.io/api-playground/api-documentation/#personal_sign, the the rpc request below, message is hex encode siwe string

```
{
    "jsonrpc": "2.0",
    "method": "personal_sign",
    "params": [ 
"0x6d796170702077616e747320796f7520746f207369676e20696e207769746820796f757220457468657265756d206163636f756e743a0a3078343333363730393963643534346339656463623266333265646333336663383061343737373734610a0a476976652074686973206170706c69636174696f6e2061636365737320746f20736f6d65206f6620796f75722064617461206f6e20436572616d69630a0a5552493a206469643a6b65793a7a364d6b726d636b6445696f6161444a5742767844667a354c38466262646b323957534e3972775863586732455553390a56657273696f6e3a20310a436861696e2049443a20350a4e6f6e63653a206145434878674e6f74790a4973737565642041743a20323032332d30322d32335432303a34373a32372e3538325a0a45787069726174696f6e2054696d653a20323032332d30332d30325432303a34373a32372e3538325a0a5265736f75726365733a0a2d20636572616d69633a2f2f2a",
        "your-eth-address"
    ],
    "id": 0
}
```

oed:
@zachferland This seems like the kind of change that would be good to test with a lot of wallets to ensure that it actually work? Only testing metamask seems a bit risky.

Can we create a list with 10 or so of the most popular wallets and test it with them?
This is the kind of thing we need to get better at testing anyway.

zachferland:
@oed dont think its necessary here because 1) it is specified as hex encoded utf8 2) major libraries like web3/ethers already convert every personal_sign message arg into hex before the provider request, ie https://github.com/web3/web3.js/blob/2b3fb3a231534e94c7b3892f37ff77100067b9e9/packages/web3-core-helpers/src/formatters.js#L221 

would assume those libraries work fine with most all providers, i would say its less likely to work with wallets in current form

oed (approved):
Ok, thats a good data point!

PR #153, opened 2023-05-11
Shared codecs package

PaulLeCam:
In line with starting to use codecs in js-ceramic, I think it would be useful to have a codecs package for the DIDs structures, that could also be used in the js-ceramic codecs package considering the overlaps for some types.

~Opening as draft to start the discussion, but will need more tests and docs.~

~@ukstv it seems there is a conflict with using both `type` and `namespace` for the `Cacao` export, any idea how to address this please? If not, I think I'll rename the type to `CacaoType` to avoid the conflict, what do you think please?~


PaulLeCam:
@zachferland if it looks OK to you do you mind if I publish the new package once the PR is merged please?
I'd like to use it in the new `@ceramicnetwork/codecs` package to avoid duplicating the codecs, cc @ukstv

packages/codecs/src/ipld.ts
```diff
@@ -0,0 +1,49 @@
+import { CID } from 'multiformats/cid'
+import { type Context, Type } from 'codeco'
+
+/**
+ * Passthrough codeco codec for CID.
+ */
+export const cid = new Type<CID, CID, unknown>(
```
zachferland, line 7: does it make sense to have the ipld ones here? didnt see where they were used

website/docs/api/classes/did_session.DIDSession.md
```diff
@@ -40,13 +40,13 @@ ___
...
...{ `s`: `TrivialCodec`<`string`\> ; `t`: `KeyOfCodec`<{ `eip1271`: ``null`` ; `eip191`: ``null`` ; `solana:ed25519`: ``null`` ; `stacks:secp256k1`: ``null`` ; `tezos:ed25519`: ``null``  }\>  }\>\>\>  }\>, `$TypeOf`\> & `MapIn`<`OptionalProps`<{ `h`: `ExactCodec`<`TypeCodec`<{ `t`: `KeyOfCodec`<{ `caip122`: ``null`` ; `eip4361`: ``null``  }\>  }\>\> ; `p`: `SparseCodec`<{ `aud`: `TrivialCodec`<`string`\> ; `domain`: `TrivialCodec`<`string`\> ; `exp`: `OptionalCodec`<`TrivialCodec`<`string`\>\> ; `iat`: `TrivialCodec`<`string`\> ; `iss`: `TrivialCodec`<`string`\> ; `nbf`: `OptionalCodec`<`TrivialCodec`<`string`\>\> ; `nonce`: `TrivialCodec`<`string`\> ; `requestId`: `OptionalCodec`<`TrivialCodec`<`string`\>\> ; `resources`: `OptionalCodec`<`Codec`<`string`[], `string`[], `unknown`\> & { `item`: `TrivialCodec`<`string`\>  }\> ; `statement`: `OptionalCodec`<`TrivialCodec`<`string`\>\> ; `version`: `TrivialCodec`<`string`\>  }\> ; `s`: `OptionalCodec`<`ExactCodec`<`TypeCodec`<{ `s`: `TrivialCodec`<`string`\> ; `t`: `KeyOfCodec`<{ `eip1271`: ``null`` ; `eip191`: ``null`` ; `solana:ed25519`: ``null`` ; `stacks:secp256k1`: ``null`` ; `tezos:ed25519`: ``null``  }\>  }\>\>\>  }\>, `$TypeOf`\>
```
zachferland, line 43: is there no nicer way to show the types in the docs? seems like it decreases readability across a lot of the docs

packages/codecs/src/did.ts
```diff
@@ -0,0 +1,29 @@
+import { refinement, string } from 'codeco'
+import type { Opaque } from 'ts-essentials'
+
+export type DIDString = Opaque<string, 'DIDString'>
```
zachferland, line 4: looks like later we could add stricter codecs for each did:key/pkh?

zachferland (commented):
is this going to change the return types for all these functions? making it a breaking change 

overall really nice have shared/defined codecs/types

packages/codecs/src/ipld.ts
zachferland: oh saw what you said before, since js-ceramic depends on dids dependencies

website/docs/api/classes/did_session.DIDSession.md
PaulLeCam: Yeah I agree it makes the generated type definitions ugly, @ukstv any idea if that could be worked around when using codeco's `TypeOf` please? Maybe using something like [`type-fest`'s `Simplify`](https://github.com/sindresorhus/type-fest/blob/main/source/simplify.d.ts)?

packages/codecs/src/ipld.ts
PaulLeCam: Yes we need it here as well to define the DagJWS codec: https://github.com/ceramicnetwork/js-did/pull/153/files#diff-381c167d850c6491abefda0be2d4d6616083316c929a147f759358602a031592R14

PaulLeCam:
Good question about the return types, I'm not sure what the behavior will be so should be safer to use major versions increases.

zachferland:
thanks for adding changesets! want to update the release section here? https://github.com/ceramicnetwork/js-did/blob/main/DEVELOPMENT.md

PaulLeCam:
> thanks for adding changesets! want to update the release section here? https://github.com/ceramicnetwork/js-did/blob/main/DEVELOPMENT.md

Yes good call, I will do it on Monday!

PaulLeCam:
@zachferland I updated the instructions, can you have a look please?

zachferland (approved):
looks good, thanks!

PR #154, opened 2023-05-11
Bump dependencies and update CI setup

PaulLeCam:
Main changes to the CI setup are dropping Node v14 and testing for v16, v18 and v20, as well as using pnpm v8.


zachferland (approved):
thanks! looks good

PR #156, opened 2023-06-13
Feat/errors helpers pkh eth

zachferland:
Two small things 
- explicit error when no app name, some people have seen this when in node and switching from web -> node auth, assuming api is same
- helper method to get accountid from did 

reminded need to add tests in most of these libraries still

### DIFF
--- a/packages/did-session/README.md
+++ b/packages/did-session/README.md
@@ -69,6 +69,17 @@ didsession.authorizations
 didsession.expiresInSecs
 ```
 
+You can also get an AccountId from a user's DID instead of the address and provider.
+
+```js
+import { DIDSession, getAccountIdByDID } from 'did-session'
+import { EthereumWebAuth } from '@didtools/pkh-ethereum'
+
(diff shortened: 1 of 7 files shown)


packages/did-session/README.md
```diff
@@ -69,6 +69,17 @@ didsession.authorizations
 didsession.expiresInSecs
 ```
 
+You can also get an AccountId from a user's DID instead of the address and provider.
+
+```js
+import { DIDSession, getAccountIdByDID } from 'did-session'
+import { EthereumWebAuth } from '@didtools/pkh-ethereum'
+
+const ethProvider = // import/get your web3 eth provider
```
oed, line 78: Might be worth mentioning https://eips.ethereum.org/EIPS/eip-1193 somewhere in our docs to be extra clear.

oed (approved):

packages/did-session/README.md
zachferland: yeah ill add that to the did-session and eth-pkh docs, thanks!

PR #163, opened 2023-08-11
chore: Do not use ethers in prod for a minor thing

ukstv:
Started as "We need to release key-did-provider-ed25519", ended with "Oh, no ethers behemoth please".

And yes, finally added changesets, so that I can release the provider.


PaulLeCam (approved):
Nice!

oed (approved):
nice

PR #164, opened 2023-08-21
Add support for persisted sessions

oed:
This PR adds support for sessions that are persisted. These sessions are created and loaded using `DIDSession.get(...)`. The sessions use the new `@didtools/key-webcrypto` did provider where keys are stored securely in the browser (outside of DOM). These keys along with the Cacao are persisted using IndexedDB.

Currently only browsers are supported. The `.get` function will throw an error if called in nodejs.

## How Has This Been Tested?

One new unit test where added that uses `fake-indexeddb` and jsdom jest env.

## Definition of Done

- [ ] Update documentation to point to this method primarily, since it's more secure

### DIFF
--- a/packages/did-session/src/index.ts
+++ b/packages/did-session/src/index.ts
@@ -277,7 +277,7 @@ function base64ToBytes(s: string): Uint8Array {
   return u8a.fromString(s, 'base64pad')
 }
 
-function cacaoContainsResources(cacao: Cacao, resources: Array<string>): boolean {
+export function cacaoContainsResources(cacao: Cacao, resources: Array<string>): boolean {
   return resources.every((res) => cacao.p.resources?.includes(res))
 }
 
@@ -335,13 +335,8 @@ export class DIDSession {
     authMethod: AuthMethod,
     authOpts: AuthOpts = {}
   ): Promise<DIDSession> {
-    if (typeof window === 'undefined')
-      throw new Error(
-        'DIDSession.get function requires browser environment, try using authorize instead'
-      )
     if (!authOpts.resources || authOpts.resources.length === 0)
       throw new Error('Required: resource argument option when authorizing')
-
     const store = await SessionStore.create()
     const result = (await store.get(account)) || {}
     let { cacao, keypair } = result as { cacao: Cacao; keypair: CryptoKeyPair }
--- a/packages/did-session/src/index.ts
+++ b/packages/did-session/src/index.ts
@@ -210,15 +210,17 @@
  */
 
 import { Ed25519Provider } from 'key-did-provider-ed25519'
+import { WebcryptoProvider, generateP256KeyPair } from '@didtools/key-webcrypto'
 import KeyDidResolver from 'key-did-resolver'
 import { randomBytes } from '@stablelib/random'
 import { DID } from 'dids'
 import { AccountId } from 'caip'
 import type { Cacao, AuthMethod, AuthMethodOpts } from '@didtools/cacao'
 import * as u8a from 'uint8arrays'
+import { SessionStore } from './sessionStore'
 
 export type SessionParams = {
-  keySeed: Uint8Array
+  keySeed?: Uint8Array
   cacao: Cacao
   did: DID
 }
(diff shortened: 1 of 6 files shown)


packages/did-session/src/index.ts
```diff
@@ -275,6 +277,10 @@ function base64ToBytes(s: string): Uint8Array {
   return u8a.fromString(s, 'base64pad')
 }
 
+function cacaoContainsResources(cacao: Cacao, resources: Array<string>): boolean {
```
PaulLeCam, line 280: Maybe this function could be exported? Seems useful for different use-cases

packages/did-session/src/index.ts
```diff
@@ -321,6 +327,45 @@ export class DIDSession {
...
 
+  /**
+   * Get a session for the given accountId, if one exists, otherwise creates a new one.
+   */
+  static async get(
+    account: AccountId,
+    authMethod: AuthMethod,
+    authOpts: AuthOpts = {}
+  ): Promise<DIDSession> {
+    if (typeof window === 'undefined')
```
PaulLeCam, line 338: Could this be moved to the `SessionStore` logic?
Would be good if we could use `globalThis` instead of `window` to access the global object, that would make the code work in workers as well.

packages/did-session/src/index.ts
```diff
@@ -321,6 +327,45 @@ export class DIDSession {
...
+      throw new Error(
+        'DIDSession.get function requires browser environment, try using authorize instead'
+      )
+    if (!authOpts.resources || authOpts.resources.length === 0)
+      throw new Error('Required: resource argument option when authorizing')
+
+    const store = await SessionStore.create()
```
PaulLeCam, line 345: Is there any risk of concurrency issue with the store getting created multiple times? Maybe the store could be a singleton?

PaulLeCam (commented):
Nice!
I'm mostly wondering if we could make it work in web workers as well?

packages/did-session/src/index.ts
oed: Hm, not sure. I think this might be handled by indexedDB?

packages/did-session/src/index.ts
oed: Good point!

packages/did-session/src/index.ts
oed: According to chatGPT, I believe we should be good assuming that two different instances of the IDBDatabase use the same lock under the hood.

#### Transactions

IndexedDB transactions can be used to help avoid race conditions. When you open a transaction, you can specify its mode as either `readonly` or `readwrite`. If a transaction is opened in `readwrite` mode, it obtains a write lock on the object store, ensuring that other `readwrite` transactions on the same object store are serialized. This can help prevent race conditions when multiple instances are writing to the database. However, read operations can still go through in parallel.

PaulLeCam: Nice, thanks for confirming!

packages/did-session/src/sessionStore.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export class SessionStore {
+  #db: IDBDatabase
+
+  constructor(db: IDBDatabase) {
+    this.#db = db
+  }
+
+  static async create(): Promise<SessionStore> {
+    if (typeof globalThis === 'undefined') {
```
PaulLeCam, line 17: I think the check should be for accessing `indexedDB`, `globalThis` should be accessible in most environments, not only browsers.

```suggestion
    if (typeof globalThis.indexedDB === 'undefined') {
```

PaulLeCam (approved):

zachferland (commented):
Looks good! 

may want to separate some of get into get/create (or other names), would allow developers to know for the UI/exp that they are asking a user to sign or not for a session

oed:
@zachferland do you think a method like `hasSessionFor(account)` would serve that purpose?

zachferland:
@oed yeah that works, `hasSessionFor(account)` and returns list with resources for each or `hasSessionFor(account, authopts/resources)` and returns bool, probably the second makes sense

PR #166, opened 2023-09-06
Add support for passkeys/webauthn

telamon:
# Add support for passkeys/webauthn
Status: `wip`
## Description
This PR adds support for using hardware tags and OS authenticators .

- added custom signer
- added custom verifier
- supports discoverable mode through key recovery

## How Has This Been Tested?

- [x] Added MockAuthenticator for Unit testing
- [x] Implemented and verified Cacao Block production through tests.

## Definition of Done

- [x] I have made corresponding changes to the documentation

## References:

[R&D Notes / Worklog](https://github.com/telamon/js-did/tree/feat/key-webauthn/packages/pkh-webauthn/journey)


packages/codecs/src/cacao.ts
```diff
@@ -47,8 +47,9 @@ export const CacaoSignature = strict(
   'CacaoSignature'
 )
 export type CacaoSignature = {
-  t: 'eip191' | 'eip1271' | 'solana:ed25519' | 'tezos:ed25519' | 'stacks:secp256k1'
+  t: 'eip191' | 'eip1271' | 'solana:ed25519' | 'tezos:ed25519' | 'stacks:secp256k1' | 'webauthn:p256'
   s: string
+  aad?: string
```
telamon, line 52: To verify a webauthn signature we need to store Additional Authenticator Data.
I added prop `aad` to `CacaoSignature` as a string was that ok?

packages/pkh-webauthn/src/utils.ts
```diff
@@ -0,0 +1,254 @@
...
+  throw new Error('Expected Uint8Array')
+}
+
+/**
+ * Tiny unsafe CBOR decoder that supports COSE_key numerical keys
+ * https://www.iana.org/assignments/cose/cose.xhtml
+ * Section 'COSE Key Type Parameters'
+ * TODO: check if iso-webauthn package handles this
+ */
+function decodeCBORHack (buf: Uint8Array) {
```
telamon, line 182: I feel a bit bad about merging awesome-sauce decoder, it does not support full CBOR spec and will throw errors if used on anything other than COSE objects.. what should I do?

telamon: On further inspection, i was wrong. package `cborg` does support negative integer keys when option `{ useMaps: true }` is passed, but in one of the testcases it lost the track of offsets and failed altogether to decode the COSE message.

https://github.com/rvagg/cborg/issues/92

There is the `iso-passkeys` library, but it uses a different eco of dependencies and attempts to support more algorithms than the `ES256 / -7`.

packages/pkh-webauthn/src/index.ts
```diff
@@ -0,0 +1,247 @@
...
+      p: {
+        domain: globalThis.location.hostname,
+        aud: '' + globalThis.location,
+        iss: '',
+        version: '1',
+        nonce: globalThis.crypto.randomUUID(),
+        resources: [],
+        exp: new Date(now + 7 * 86400000).toISOString(), // 1 week
+        nbf: new Date(now).toISOString(),
+        iat: new Date(now).toISOString(),
```
oed, line 173: Please default to opts:
https://github.com/ceramicnetwork/js-did/blob/main/packages/cacao/src/cacao.ts#L96-L102

packages/pkh-webauthn/src/index.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    const block = await blockFromCacao(challenge) // await CacaoBlock.fromCacao(challenge) when issue resolved
+    // perform sign
+    const { credential, recovered } = await authenticatorSign(block.cid.bytes)
+
+    // @ts-ignore - credential.response does exist.
+    const { response } = credential
+    const { clientDataJSON, signature } = response
+    const authData = getAuthenticatorData(response)
+    const aad = u8a.toString(assertU8(encode({ authData, clientDataJSON })), 'base64url')
```
oed, line 186: Can we just store this as a byte array (Uint8Array) instead of a string please.

packages/pkh-webauthn/src/index.ts
```diff
@@ -0,0 +1,247 @@
...
+    for (const selector of session.selectors) {
+      if (typeof selector.select !== 'function') continue
+      if (pk = await selector.select(credential.id, ...recovered)) break
+    }
+    if (!pk) throw new Error('PublicKeySelectionFailed')
+    if (!recovered.find(c => !u8a.compare(c, pk))) throw new Error('UnrelatedPublickey')
+
+    // Insert iss + aad after signature
+    return {
+      ...challenge,
```
oed, line 198:
```suggestion
      h: challenge.h,
```

Should be sufficient.

oed (changes requested):
A first pass going over some quick things that needs to be addressed.

I also feel like we could potentially simplify the API and leave what you are calling "sessions" out of scope of this API and up to consumers.

### This is what I'm thinking:

First off we always create discoverable credentials. Then we have the following methods. Seems like you were going mostly in this direction, so would be good to discuss tradeoffs here.

```ts
// creates a new discoverable credential and converts the public key into a `did:key` and returns it.
createDID(): string

// signs one message using discoverable credentials and returns both possible public keys encoded as DIDs.
probeDIDs(): Array<string>

// takes either a DID directly (from `createDID` method, or from developer stored wherever they prefer), or two DIDs from `probeDIDs`.
getAuthMethod({ did: string, oneOfDids: Array<string> }): AuthMethod

File: packages/pkh-webauthn/src/index.ts [hunk 6]
```diff
@@ -0,0 +1,247 @@
...
+    // The public key is not known at pre-sign time.
+    // so we sign a "challenge"-block without Issuer attribute
+    const challenge: Cacao = Object.freeze({
+      h: {
+        t: 'caip122'
+      },
+      p: {
+        domain: globalThis.location.hostname,
+        aud: '' + globalThis.location,
+        iss: '',
```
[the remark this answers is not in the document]
telamon: yes i wasn't sure if the encoder was stable, removing sanity-seat-belt =)

packages/pkh-webauthn/src/index.ts
telamon: Yes. I zoomed past the entire options object i see. thanks for the link.

telamon:
> A first pass going over some quick things that needs to be addressed.
> 
> I also feel like we could potentially simplify the API and leave what you are calling "sessions" out of scope of this API and up to consumers.
> ### This is what I'm thinking:
> 
> First off we always create discoverable credentials. Then we have the following methods. Seems like you were going mostly in this direction, so would be good to discuss tradeoffs here.
> 
> ```ts
> // creates a new discoverable credential and converts the public key into a `did:key` and returns it.
> createDID(): string
> 
> // signs one message using discoverable credentials and returns both possible public keys encoded as DIDs.
> probeDIDs(): Array<string>
> 
> // takes either a DID directly (from `createDID` method, or from developer stored wherever they prefer), or two DIDs from `probeDIDs`.
> getAuthMethod({ did: string, oneOfDids: Array<string> }): AuthMethod
> ```

Glad you brought this up.  
I attempted to save adopters from the headache of _when_ to probe by doing it "on-demand"
but after closer inspection it seems I just caused the headache elsewhere.

The methods `createDID()` and `probeDIDs()` are straightforward.  
The difference is that the DID becomes settled/known after `AuthMethod` was used instead of after the probe.
The end result is identical to original implementation but at a lower cost of complexity and a clearer API.
(I think yes :+1: )

I considered the following scenarios:
|authenticator state|browser state|user-choice|Action|button presses|Outcome|
|---|---|---|---|--:|---|
|*|unknown|New ID|`createDID(); getAuthMethod(did)`|2|success|
|has key|known|New ID|`forget();createDID(); getAuthMethod(did)`|2|success|
|has key|known|Login|`getAuthMethod(did)`|1|success|
|has key|unknown|Login|`probeDIDs(); getAuthMethod(candidates)`|2|success|
|empty|unknown|Login|`probeDIDs();`|1|fail|

### API Proposal
Given the new methods; I would update the example in the `README.md` to:
```js
// New identity
const did = await createDID('label')
await storeSomehow(did)
const authMethod = await getAuthMethod({ did })

// Returning user
const did = await getSomehow()
let authMethod

if (did) { // previous DID found
  authMethod = await getAuthMethod({ did })
} else { // new device detected
  const dids = await probeDIDs()
  authMethod = await getAuthMethod({ dids })
}

const session = await DIDSession.authorzie(authMethod, { resources: [...] })
await storeSomehow(session.did.id) // Save for later
```

Is that ok?

oed:
Thanks for the suggested api. A few questions/notes:

1. in `createDID('label')`, what is the purpose of the parameter?
2. Maybe make a comment about what the developer should do rather than include methods names not exported by this library.
3. You need to store `session.did.parent`, not `session.did.id` (the latter is just the session key)

telamon:
@oed 2,3 gotcha! 

1: The label is the name of the credential, it's what's displayed in OS-popups on discoverable ops:
image


#### different question
if probe returns two fully qualified did-strings, does that mean we export one false DID?
Not sure about the vocabulary but gut tells me to call it a recoveryToken or probeResult to avoid accidental storage/transmission:
```js
const recoveryToken = await probe()

const authMethod = getAuthMethod({ did, recoveryToken })
```
Is this a concern?

oed:
1 . Makes sense. What happens if there are two keys with the same label?

4 . Ok, makes sense to call it something like `recoveryToken` as you suggest.

5 . I wonder if we should also add another option in `getAuthMethod` that would allow you to pass a function as follows:
```js

function selectDID(did1: string , did2: string): Promise<did>

const authMethod = getAuthMethod({ selectDID })
```
Calling `getAuthMethod` would result in one signature to sign the CACAO, then the user could resolve the correct DID in whichever way they want.

telamon:
1. They are more or less shown as dupes (isolated to domain):
image

(webauthn does not export functions to remove/relabel credentials, so we can't really do much more here.)


5. The pure functional probe approach you suggested gives a solid API and provides developers freedom in implementation.  
Using a select callback is valid, but let it be known that I went down that path by accident.. (I failed to recognize that the order of `probe()` and `sign()` does not matter for recovery). 
The cons of conditional callbacks is that they are more difficult to debug/trigger, and affects only those who wish to use passkeys in a decentralized manner.

Thinking "There should be one clear way to do it.": 
I cast my vote on the manual probe instead of callback; (makes more sense, causes less developer strain *shrug*)

oed:
I would prefer if we could do both probe and callback approaches.

For example in Ceramic we could store a document after the first time we authenticate. This would allow a developer to query documents for the two DIDs and if it finds one of them it knows which is the correct DID.

packages/pkh-webauthn/src/index.ts
```diff
@@ -0,0 +1,221 @@
...
+    // Insert iss + aad after signature
+    return {
+      h: challenge.h,
+      p: {
+        ...challenge.p,
+        iss,
+      },
+      s: {
+        t: 'webauthn:p256',
+        s: u8a.toString(signature, 'base64url'),
```
oed, line 127: `s` should be Bytes, e.g. Uint8Array, according to [CACAO spec](https://chainagnostic.org/CAIPs/caip-74#specification)
```suggestion
        s: signature,
```

packages/pkh-webauthn/src/index.ts
```diff
@@ -0,0 +1,221 @@
...
+        aad
```
oed, line 128: I think that any additional data should be part of `m` (SignatureMeta struct).
```suggestion
        m: { aad }
```

packages/pkh-webauthn/src/utils.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  // https://w3c.github.io/webauthn/#sctn-attested-credential-data
+  const aaguid = authData.slice(o, o += 16)
+  const clen = view.getUint16(o); o += 2
+  const credentialId = authData.slice(o, o += clen)
+
+  // https://datatracker.ietf.org/doc/html/rfc9052#section-7
+
+  // Bug in cborg.decode() prevents next-line: https://github.com/rvagg/cborg/issues/92
+  // const cose = decode(authData.slice(o - 1), { useMaps: true }) as Map<number, number|Uint8Array>
```
oed, line 90: Was this resolved in `cborg` 4.0.0?

packages/pkh-webauthn/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "@didtools/pkh-webauthn",
```
oed, line 2:
```suggestion
  "name": "@didtools/key-webauthn",
```

Let's call this package `key-webuathn` since it's definitely not using the `pkh` DID.

oed (changes requested):
Looks really great. Just a few changes required from my end.

ukstv (approved):

packages/pkh-webauthn/src/utils.ts
telamon: Yup, upgraded, tested, works!

packages/pkh-webauthn/src/index.ts
telamon: I see, the type-definitions seem to be outdated compared to spec. `codecs/src/cacao.ts:51`.

I don't want to update the typedefs in this PR as it will most likely break static-analysis for modules that I am not familiar with. 
I think updating typedefs belongs in a separate issue; Changes applied but leaving a `@ts-ignore + TODO:` in `test/index.test.ts` for now.

oed: cc @ukstv Thoughts about tyhpedefs?

oed: @ukstv so the spec is wrong or we are wrong?

ukstv: @oed we are wrong.

oed: ugh, ok. This might be a problem if WalletConnect implements CAIP-222.

>> Looks closer

Actually in [CAIP-22 a string is also used](https://chainagnostic.org/CAIPs/caip-222#response), but it is a hex string 🙃

ukstv: Solana: base58btc-encoded, Ethereum: 0xHex, Tezos: tezos-specific signature string, that's what we have now.

ukstv: So, base64url (no pad) is okayish.

oed: Ok, let's use the base64url string then.

telamon: @ukstv , @oed , Sorry for the delay, I needed to spend some time offline. 

I've patched the following:
- Reverted signature encoding back to base64url, tests pass.
- Added usage instruction to `src/indext.ts` following codebase praxis.

let me know if there is anything else i can do, otherwise hoping for a merge. Thanks

oed (commented):
This should be good to merge once tests pass!

oed (approved):

PR #171, opened 2023-10-07
feat: Use threads for jws verification

dbcfd:


packages/dids/src/did.ts
```diff
@@ -267,29 +286,12 @@ export class DID {
...
-        )
-      }
-      options.protected = options.protected || {}
-      options.protected.cap = `ipfs://${capCID?.toString()}`
-    }
-    const { jws } = await this._client.request('did_createJWS', {
-      did: this._id,
-      ...options,
-      payload,
+    return await createJWSUsing({
```
ukstv, line 289: You do not have to do `return await` here. And, `createJWS` could be unmarked `async`. Wrapping/unwrapping promises adds an overhead. Same for other `return await` clauses.

ukstv (approved):
Approved, conditional to `return await` changes. It is not a big deal. Just a little bit more sanity.

PR #175, opened 2023-12-13
key-webauthn Compatibility chart and Bugfix

telamon:
# key-webauthn Compatibility chart and Bugfix

## Description
In this PR i've added a demo page that uses `@didtools/key-webauthn` in runtime.
this led to the discovery of a couple of bugs and the production of a browser/device/authenticator chart.

- Fixed ArrayBuffer => Uint8Array type mismatch in runtime
- Added Compatibility section to README.md
- Created a login-demo in subfolder ./sample

Just to clarify the exported `key-webauthn` package was mostly left unchanged with the exception
of aformentioned type error.

## How Has This Been Tested?

The [demo](https://didtoolswn.surge.sh/) was manually tested on multiple browsers and devices.

## Definition of Done

- [x] I have made corresponding changes to the documentation


oed (approved):
lgtm

ukstv (approved):

ukstv:
It is a bit ugly in a sense, that it relies on unpkg and does not work offline,  - a bundler might be handy here. And Tonic - not a battle tested React. We'll have to maintain this non-prod thing, and React is more familiar. Otherwise okay.

telamon:
> It is a bit ugly in a sense, that it relies on unpkg and does not work offline, - a bundler might be handy here. And Tonic - not a battle tested React. We'll have to maintain this non-prod thing, and React is more familiar. Otherwise okay.

@ukstv Yes i wrote a non-prod tool to test production code in runtime, I saw value in committing the source so the results could be reproduced, but I did not wish to pollute the dependency tree hence unkpkg.
I used tonic because it's fun; ~400LOC; Native templates; no event loop; no virtual-dom.

PR #176, opened 2023-12-13
Support eip55

### LINKED ISSUE #173: The signature problem on Integrating with Lit
# Description

I'm trying to approve for Lit Protocol and Ceramic with just one signature. I've managed to do so, but there might be some issues with my solution. I'll describe my problems here and discuss if there's a better approach, or whether it's possible to optimize the code of Lit or Ceramic to better address these issues.

# Technical Information

My approach is as follows: first, I call Ceramic's authorization, with the code as below:
``` TypeScript
const accountId = await getAccountId(ethProvider, address)
      
const authMethod = await getAuthMethod(ethProvider, accountId)
console.log("[AuthCeramic] Authorize", authMethod)

session = await DIDSession.authorize(authMethod, {resources: ["ceramic://*"]})
```
After the user signs, they get a session as a login credential.
In Lit Protocol, similar to session, there's a data structure called AuthSig as a login credential. The structure of AuthSig is as follows:
``` TypeScript
interface AuthSig {
    sig: any; // Data after signing
    derivedVia: string; // Generally "web3.eth.personal.sign"
    signedMessage: string; // Data before signing
    address: string; // Signing address
}
```

Actually, session and AuthSig are somewhat interchangeable to a certain extent. Below is my conversion code:
``` TypeScript
async function getAuthSig(session, chainId = TargetChainId) {
  const {p, s} = session.cacao;
  const address = p.iss.split(":").pop();

  const message = new SiweMessage({
    domain: p.domain, address, 
    statement: p.statement,
    uri: p.aud,
    version: p.version,
    nonce: p.nonce,
    issuedAt: p.iat,
    expirationTime: p.exp,
    chainId: chainId.toString(),
    resources: p.resources,
  })

  return {
    address, derivedVia: "web3.eth.personal.sign",
    sig: s.s, signedMessage: message.toMessage(),
  } as AuthSig
}
```
Therefore, we can convert the session obtained after authorizing Ceramic into AuthSig and set it in LocalStorage, to authorize Lit:
``` TypeScript
const LitAuthSigKey = "lit-auth-signature"
const LitProviderKey = "lit-web3-provider"
const LitDefaultProvider = "metamask"

if (authSig) { // Set the default authSig in advance to avoid reconnecting and re-signing
  setLocalStorage(LitProviderKey, LitDefaultProvider)
  setLocalStorage(LitAuthSigKey, authSig)
}
authSig = await LitJsSdk.checkAndSignAuthMessage({
  chain: Chain
});

// Do some encryption/decryptions ....
const res = await LitJsSdk.decryptToString({
  authSig, chain: Chain, ...
}, lit)
```
This is my approach.

# Solution Problem

There's a problem with the above solution, which is the issue of case sensitivity in the address in the signature information.

In the signature information of Ceramic, the address is converted to lowercase, here's an example:
```
localhost wants you to sign in with your Ethereum account:
0x66d369e52b1a7ee16.......f84025

Give this application access to some of your data on Ceramic

URI: did:key:z6Mkofcp......pnsDJ26j
Version: 1
Chain ID: 1
Nonce: 1CjSZZjMv2
Issued At: 2023-11-15T08:13:19.690Z
Expiration Time: 2023-11-22T08:13:19.690Z
Resources:
- ceramic://*
```
However, when we convert this signature into AuthSig and authenticate it in Lit, it says that the address does not conform to the EIP-55 format:
a1fbbbbb99c57d8ab7f2f9a94c12187
And this is the request data:
image

Therefore, I have to insert the following statement in the first section of the code to ensure that the address in the signature information is in EIP-55 format:
``` TypeScript
const accountId = await getAccountId(ethProvider, address)
accountId.address = address // Prevent the lowercase address
```
Only then can I successfully authorize Lit.
But after converting to EIP-55, the generated PKH-DID looks like this: did:pkh:eip155:1:0x66d369E52b1A7ee16D65e6a052745561C1f84025, which contains case sensitivity, and that's the problem:
- If I keep the address lowercase, the generated DID is normal and conforms to the rules, but Lit cannot be successfully authorized.
- If I retain the case sensitivity, I can successfully authorize Lit, but the generated DID will also retain the case sensitivity, but it seems to be able to be used normally in Ceramic.

Therefore, I wonder if there is a better solution, or whether it's possible to make some modifications to Ceramic or Lit to solve this problem?

oed:
This PR enables the pkh-ethereum CACAO verifier to verify SIWE messages that have the ethereum address in the `iss` field encoded using EIP-55, while retaining support for lower case addresses.

In addition it would be desirable to change the default behavior of generated SIWE messages for CACAO to use eip55 by default since the spec says the address ["Must" be encoded using eip55](https://eips.ethereum.org/EIPS/eip-4361#message-format). However, since the `did-session` package uses `@ceramicnetwork` packages that depend on a released version of `@didtools` it's not possible to make this change now before the verification support has been rolled out.

This PR fixes #173


packages/cacao/package.json
```diff
@@ -48,7 +48,8 @@
     "@ipld/dag-cbor": "^9.0.6",
     "caip": "^1.1.0",
     "multiformats": "^12.1.3",
-    "uint8arrays": "^4.0.9"
+    "uint8arrays": "^4.0.9",
+    "viem": "^1.19.13"
```
PaulLeCam, line 52: Any way to avoid importing this full library for just one utility function please? Maybe there are alternative packages that would be lighter?

packages/cacao/src/siwx/siwe.ts
```diff
@@ -13,22 +14,31 @@ export class SiweMessage extends SiwxMessage {
     return super.toMessage('Ethereum')
   }
 
+  toMessageEip55(): string {
+    const tmpAddress = this.address
+    this.address = checksumAddress(this.address as `0x${string}`)
+    const msg = super.toMessage('Ethereum')
+    this.address = tmpAddress
```
PaulLeCam, line 21: Seems pretty hacky to have to change the address on the instance twice like this to make it work, could the address be provided as extra argument to the `toMessage()` method instead?

packages/cacao/src/siwx/siwe.ts
```diff
@@ -13,22 +14,31 @@ export class SiweMessage extends SiwxMessage {
...
+  }
+
   /**
    * This method parses all the fields in the object and creates a sign
    * message according with the type defined.
    * @returns {string} Returns a message ready to be signed according with the
    * type defined in the object.
    */
-  signMessage(): string {
+  signMessage(eip55?: boolean): string {
```
PaulLeCam, line 31: Could we use an options object here to be more explicit and in case we need to add more options in the future?
```suggestion
  signMessage(options: {eip55?: boolean} = {}): string {
```

packages/cacao/package.json
oed: Viem is very likely going to be a dependency for the upcoming [Varsig implementation](https://github.com/ceramicnetwork/js-did/blob/feat/varsig/packages/varsig/package.json#L66), which means it's pretty much always going to be a requirement everywhere. 

I think viem is pretty tree-shakable, so it shouldn't have much of an impact on actual build sizes?

PaulLeCam: OK thanks, sounds good if tree-shaking works well 👍

PaulLeCam (approved):

packages/cacao/package.json
oed: Well that's what I see on https://bundlephobia.com/package/viem

Can I trust the badge shown there?

PaulLeCam: I don't know how reliable it is, I guess it depends on our own build process as well?
Would be good to figure out though, the lib is 200kB+ so a pretty big addition to a low-level library.

oed: Any thoughts on this @ukstv

PR #183, opened 2024-01-04
chore: multifomats 13 for codecs

dbcfd:
Version bump for codecs so it can release. Fix repository in pkh-did-resolver so that it can be publishshed.

codecs@3.0.0


dav1do (approved):
Are we having fun yet!?

PR #184, opened 2024-01-04
fix(dids): Verify eip55 formatted CACAOs correctly

oed:
There was a bug introduced in #176 where CACAO/SIWE message signed with eip55 encoding would result in an  `unsupportedDidMethod` error. This was because an equality check would fail due to the difference between a lower case and a mixed case (eip55) ethereum address. This PR fixes this problem.

### DIFF
--- a/packages/dids/src/did.ts
+++ b/packages/dids/src/did.ts
@@ -113,6 +113,14 @@ function isResolver(resolver: Resolver | ResolverRegistry): resolver is Resolver
   return 'registry' in resolver && 'cache' in resolver
 }
 
+function issuerEquals(dida: string, didb: string): boolean {
+  if (dida === didb) return true
+  if (dida.startsWith('did:pkh:eip155:1:')) {
+    return dida.toLowerCase() === didb.toLowerCase()
+  }
+  return false
+}
+
 /**
  * Interact with DIDs.
  */
@@ -368,7 +376,7 @@ export class DID {
     const signerDid = didResolutionResult.didDocument?.id
     if (
       options.issuer &&
-      options.issuer === options.capability?.p.iss &&
+      issuerEquals(options.issuer, options.capability?.p.iss) &&
       signerDid === options.capability.p.aud
     ) {
       if (!options.verifiers) throw new Error('Registered verifiers needed for CACAO')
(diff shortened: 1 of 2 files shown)


packages/dids/src/did.ts
```diff
@@ -113,6 +113,14 @@ function isResolver(resolver: Resolver | ResolverRegistry): resolver is Resolver
   return 'registry' in resolver && 'cache' in resolver
 }
 
+function issuerEquals(dida: string, didb: string): boolean {
```
dav1do, line 116: Nice, it looks like this fixes the bug but there might be something else. This test previously had the following code and passed. Is it expected/desired to change? We get [received like this](https://github.com/ceramicnetwork/js-ceramic/blob/17d77ca29b8c52e24750b5906107ccb9375438e7/packages/stream-model-instance/src/model-instance-document.ts#L309C54-L309C54) (signer.did.parent or signer.did.id)

```
  ● CACAO Integration test › Ceramic dids instance with capability/parent › can create and update new model stream with model resource

    expect(received).toEqual(expected) // deep equality

    Expected: "did:pkh:eip155:1:0xBd9D9c7DC389715a89fC8149E4a5Be91336B2796"
    Received: "did:pkh:eip155:1:0xbd9d9c7dc389715a89fc8149e4a5be91336b2796"

      454 |
      455 |       expect(doc.content).toEqual(CONTENT0)
    > 456 |       expect(doc.metadata.controller).toEqual(`did:pkh:eip155:1:${wallet.address}`)
          |                                       ^
      457 |       expect(doc.metadata.model.toString()).toEqual(METADATA.model.toString())
      458 |
      459 |       await doc.replace(CONTENT1, {

      at Object.toEqual (src/__tests__/capability.test.ts:456:39)

```

oed: This is an error in the tests of js-ceramic. We should always expect the `controller` of DID PKH in Ceramic to be lowercase if it's `eip155`

dav1do: Okay, fair enough. I'm still not totally sure why it wasn't lowercase before but it's easy to adjust the test.

dav1do (approved):

PR #187, opened 2024-02-12
feat(did-session): session removal support

oed:
## Description
This PR adds support for removing sessions that have been created using `DIDSession.get(...)`. 

Additionally:
- `DIDSession.hasSessionFor(...)` now returns false if the persisted session is expired
- Added tests to make sure sessions are properly renewed if expired on call to `DIDSession.get(...)`

## How Has This Been Tested?

Cleaned up the test suite for DIDSession package into categories and added new tests for persisted sessions.


ukstv (approved):
wow

PR #193, opened 2024-07-12
fix: make SIWE default to EIP55

oed:
A while back we updated the `@didtools/cacao` package to be able to verify SIWE messages with both eip55 encoded ethereum addresses as well as lower case ones. This is because the SIWE spec requires the address to be eip55. We needed to roll this change out to `@ceramicnetwork/*` packages before we could change the default behavior to sign eip55 messages from the pkh-ethereum package. This PR makes this change.

### DIFF
--- a/packages/pkh-ethereum/src/authmethod.ts
+++ b/packages/pkh-ethereum/src/authmethod.ts
@@ -78,7 +78,7 @@ async function createCACAO(
     resources: opts.resources,
   })
   const signature = await safeSend(ethProvider, 'personal_sign', [
-    encodeHexStr(siweMessage.signMessage()),
+    encodeHexStr(siweMessage.signMessage({ eip55: true })),
     normAccount.address,
   ])
   siweMessage.signature = signature
--- a/packages/did-session/package.json
+++ b/packages/did-session/package.json
@@ -49,11 +49,11 @@
     "uint8arrays": "^5.0.1"
   },
   "devDependencies": {
-    "@ceramicnetwork/common": "^3.2.0",
-    "@ceramicnetwork/stream-model": "^2.2.0",
-    "@ceramicnetwork/stream-model-instance": "^2.2.0",
-    "@ceramicnetwork/stream-tile": "^3.2.0",
-    "@ceramicnetwork/streamid": "^3.2.0",
+    "@ceramicnetwork/common": "^5.15.0",
+    "@ceramicnetwork/stream-model": "^4.15.0",
+    "@ceramicnetwork/stream-model-instance": "^4.16.0",
+    "@ceramicnetwork/stream-tile": "^5.15.0",
+    "@ceramicnetwork/streamid": "^5.4.0",
     "@didtools/cacao": "workspace:^",
     "@didtools/pkh-ethereum": "workspace:^",
     "@didtools/pkh-solana": "workspace:^",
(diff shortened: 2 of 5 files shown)


dav1do (approved):

packages/pkh-ethereum/src/authmethod.ts
```diff
@@ -78,7 +78,7 @@ async function createCACAO(
     resources: opts.resources,
   })
   const signature = await safeSend(ethProvider, 'personal_sign', [
-    encodeHexStr(siweMessage.signMessage()),
+    encodeHexStr(siweMessage.signMessage({ eip55: true })),
```
derekpierre, line 81: I thought line https://github.com/ceramicnetwork/js-did/pull/193/files#diff-d2c60f04e1dcb4772ca82cf036d7f3826d992a107c54b8ff1e844128fb091189R99 was the issue with the casing of the address?

What you are signing might be modified to be eip55 but the message itself doesn't seem to have the casing.

Or am I mistaken?

derekpierre: Also this - https://github.com/ceramicnetwork/js-did/blob/main/packages/pkh-ethereum/src/authmethod.ts#L66C9-L66C20...?

oed: So this get's a bit complicated because Ceramic needs the DID in lowercase format for the `controller` of any given stream. 

The `.signMessage` is badly named. It really just returns the messaged to be sent to the wallet and signed. You can see the implementation accepts the eip55 parameter here:
https://github.com/ceramicnetwork/js-did/blob/3a8ad6d2c743b9b22c067961ae0e71818ff76443/packages/cacao/src/siwx/siwe.ts#L28-L43

derekpierre: Ok. So I would use toMessageEip55() to get the string with the correct casing.

The signature associated with the SiweMessage object used for the session, did that use lower case for the address when signing the string, or does it use the EIP-55 format?

oed: Yes, and `toMessageEip55()` in turn uses `checksumAddress` from the `viem` package.

The string that gets sent to the wallet for signing should be a valid SIWE message, i.e. use eip55 encoding for the address.

derekpierre: I switched to using `toMessageEip55()` but now there is a signature mismatch with the signature on the SIWE message object.

Here is the code I use to get the information from the DID session:

```typescript
const session = await DIDSession.fromSession(sessionStr);
const message = SiweMessage.fromCacao(session.cacao);
const messageStr = message.toMessageEip55();
const signature = message.signature;
```

Later on when validating the signature with the message, the validation fails:
```
error:  SiweError
expected: "0xf9B4Fbf41C540ba28491a5B9EFD9BbcaBaaa95d3"
received: "Resolved address to be 0x2d1F436Ce1A7Ca3959f23D90981B2e65f46629e6"
type: "Signature does not match address of the message."
```

oed: @derekpierre Why are you calling `.toMessageEip55()` separately? Maybe if you want to share more of your code I can help to see what's going wrong?

derekpierre: Ah, it was an old dependency. It works now! 

I do need to call `.toMessageEip55()` because I'm trying to reuse the same SIWE message and signature that Ceramic stores in the DIDSession. This message/signature combination is passed to TACo nodes as proof of ownership of a wallet - so instead of generating a separate one, we just reuse Ceramic's as a single sign-on type of functionality.

